PR #2, opened 2022-01-06
Copy NOMIS User Roles API

petergphillips:
- ✨ Initial stab at copying from nomis user roles api
- ✨ Temporarily build on branch to get deployment out


helm_deploy/hmpps-nomis-prisoner-api/Chart.yaml
```diff
@@ -7,6 +7,3 @@ dependencies:
   - name: generic-service
     version: 1.0.16
     repository: https://ministryofjustice.github.io/hmpps-helm-charts
-  - name: generic-prometheus-alerts
```
petergphillips, line 10: not needed as deploying to digital studio aks

Mjwillis (approved):
LGTM

PR #9, opened 2022-01-12
Initial version of JPA mappings for visits

steverendell:


Mjwillis (approved):
This all looks very familiar !  Nice to see it in Kotlin

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderBooking.kt
```diff
@@ -0,0 +1,208 @@
...
+  @Column(name = "DISCLOSURE_FLAG", nullable = false)
+  val disclosureFlag: String = "Y",
+
+  @Column(name = "BOOKING_BEGIN_DATE", nullable = false)
+  val bookingBeginDate: LocalDateTime,
+
+  //
+  //    @OneToMany(mappedBy = "id.offenderBooking", cascade = CascadeType.ALL)
+  //    @Default
+  //    private List<OffenderProfileDetail> profileDetails = new ArrayList<>();
```
Mjwillis, line 50: Wonder if we'll ever be able factor this common data into a library of some kind?

Mjwillis (approved):
LGTM

PR #12, opened 2022-01-17
SDI-15 Create visit nomis endpoint

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/data/AmendVisitRequest.kt
```diff
@@ -0,0 +1,35 @@
+package uk.gov.justice.digital.hmpps.nomisprisonerapi.data
+
+import com.fasterxml.jackson.annotation.JsonInclude
+import io.swagger.v3.oas.annotations.media.Schema
+import java.time.LocalDateTime
+import javax.validation.constraints.NotBlank
+import javax.validation.constraints.NotEmpty
+import javax.validation.constraints.Size
+
```
steverendell, line 9: amend and cancel just placeholders for now

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/VisitVisitor.kt
```diff
@@ -27,16 +27,15 @@ data class VisitVisitor(
   @Column(name = "OFFENDER_VISIT_VISITOR_ID", nullable = false)
   val id: Long? = null,
```
steverendell, line 29: mapping the id makes more sense in an insert scenario

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/resource/VisitResource.kt
```diff
@@ -0,0 +1,154 @@
...
+            mediaType = "application/json",
+            schema = Schema(implementation = ErrorResponse::class)
+          )
+        ]
+      ),
+    ]
+  )
+  fun createVisit(@RequestBody @Valid createVisitRequest: CreateVisitRequest): CreateVisitResponse =
+    visitService.createVisit(createVisitRequest)
+
```
steverendell, line 76: amend and cancel todo - just a discussion point really re what the endpoint should look like

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitService.kt
```diff
@@ -0,0 +1,169 @@
...
+
+    val offenderBooking = offenderBookingRepository.findByOffenderNomsIdAndActive(visitDto.offenderNo, true)
+      .orElseThrow(PrisonerNotFoundException(visitDto.offenderNo))
+
+    createBalance(visitDto, offenderBooking)
+
+    val visit = visitRepository.save(mapVisitModel(visitDto, offenderBooking))
+
+    addVisitors(visit, offenderBooking, visitDto)
+
```
steverendell, line 59: assumed an event would be useful

src/main/resources/db/migration/data/R__2_3__REFERENCE_CODES.sql
```diff
@@ -2732,6 +2732,7 @@ INSERT INTO REFERENCE_CODES (DOMAIN, CODE, PARENT_DOMAIN, PARENT_CODE, DESCRIPTI
 INSERT INTO REFERENCE_CODES (DOMAIN, CODE, PARENT_DOMAIN, PARENT_CODE, DESCRIPTION, LIST_SEQ, ACTIVE_FLAG, SYSTEM_DATA_FLAG, EXPIRED_DATE) VALUES ('VIS_COMPLETE', 'OFFEND', null, null, 'Offender Completed Early', 99, 'Y', 'N', null);
 INSERT INTO REFERENCE_CODES (DOMAIN, CODE, PARENT_DOMAIN, PARENT_CODE, DESCRIPTION, LIST_SEQ, ACTIVE_FLAG, SYSTEM_DATA_FLAG, EXPIRED_DATE) VALUES ('VIS_COMPLETE', 'SCH', null, null, 'Scheduled', 99, 'Y', 'Y', null);
 INSERT INTO REFERENCE_CODES (DOMAIN, CODE, PARENT_DOMAIN, PARENT_CODE, DESCRIPTION, LIST_SEQ, ACTIVE_FLAG, SYSTEM_DATA_FLAG, EXPIRED_DATE) VALUES ('VIS_COMPLETE', 'VISITOR', null, null, 'Visitor Completed Early', 99, 'Y', 'N', null);
```
steverendell, line 2734: from Petes update PR

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/VisitVisitor.kt
andymarke: Do you mean `personId` ? Wouldn't you look up Person anyway to verify the personId passed in is correct?

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/resource/VisitResource.kt
```diff
@@ -0,0 +1,154 @@
...
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.data.CancelVisitRequest
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.data.CreateVisitRequest
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.data.CreateVisitResponse
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.service.VisitService
+import javax.validation.Valid
+
+@RestController
+@Validated
+@RequestMapping("/visits", produces = [MediaType.APPLICATION_JSON_VALUE])
+class VisitResource(private val visitService: VisitService) {
```
andymarke, line 30: I was wondering whether these should be resources of a prisoner?
e.g. it would be 

`/prisoner/offenderNo/{offenderNo}/visit`

So you must supply a prisoner id separately from the body?

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/resource/VisitResource.kt
```diff
@@ -0,0 +1,154 @@
...
+
+  @PreAuthorize("hasRole('ROLE_UPDATE_NOMIS')")
+  @PostMapping
+  @ResponseStatus(HttpStatus.CREATED)
+  @Operation(
+    summary = "Creates a new visit",
+    description = "Creates a new visit and decrements the visit balance.",
+    security = [SecurityRequirement(name = "ROLE_UPDATE_NOMIS")],
```
andymarke, line 38: I can't remember, does this SecurityRequirement annotation do anything? Does it come out in the swagger?

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitService.kt
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  fun cancelVisit(visitId: Long, visitDto: CancelVisitRequest) {
+  }
+
+  private fun createBalance(
+    visitDto: CreateVisitRequest,
+    offenderBooking: OffenderBooking,
+  ) {
+    val offenderVisitBalance = offenderVisitBalanceRepository.findById(offenderBooking.bookingId!!).orElseThrow()
```
andymarke, line 75: `bookingId!!`  implies it is optional? Is could we make it not optional or use `lateinit` it feels there should be away around describing an id as mandatory but it is set as soon as the entity is created.
Not sure what we have done elsewhere

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitService.kt
```diff
@@ -0,0 +1,169 @@
...
+        eventStatus = scheduledEventStatus,
+      )
+    )
+
+    visitDto.visitorPersonIds.forEach {
+      visitVisitorRepository.save(
+        VisitVisitor(
+          visitId = visit.id,
+          personId = it,
+          eventStatus = scheduledEventStatus,
```
andymarke, line 158: feels like this should be a `person` not `personId` else couldn't pass any number you like and corrupt NOMIS?

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitService.kt
```diff
@@ -0,0 +1,169 @@
...
+    visitDto: CreateVisitRequest
+  ) {
+    val scheduledEventStatus = eventStatusRepository.findById(SCHEDULED_APPROVED).orElseThrow()
+
+    //  Add dummy visitor row for the offender_booking as is required by the P-Nomis view
+
+    visitVisitorRepository.save(
+      VisitVisitor(
+        visitId = visit.id!!,
+        offenderBooking = offenderBooking,
```
andymarke, line 148: maybe this should be mapped to `visit` rather than `visitId` - again feels weird you need to `visit.id!!`

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/VisitVisitor.kt
```diff
@@ -27,16 +27,15 @@ data class VisitVisitor(
...
+  val offenderBooking: OffenderBooking? = null,
 
   @Column(name = "OFFENDER_VISIT_ID", nullable = false)
   val visitId: Long,
 
-  @ManyToOne(optional = false, fetch = FetchType.LAZY)
-  @JoinColumn(name = "PERSON_ID", nullable = false)
-  val person: Person,
+  @Column(nullable = false)
+  val personId: Long? = null,
```
lbennett-moj, line 39: is this for the dummy row scenario with offender booking?

steverendell: No its for all - dummy or visitors. Ive not 'hidden' that or done anything clever like filters etc

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitService.kt
```diff
@@ -0,0 +1,169 @@
...
+  fun createVisit(visitDto: CreateVisitRequest): CreateVisitResponse {
+
+    val offenderBooking = offenderBookingRepository.findByOffenderNomsIdAndActive(visitDto.offenderNo, true)
+      .orElseThrow(PrisonerNotFoundException(visitDto.offenderNo))
+
+    createBalance(visitDto, offenderBooking)
+
+    val visit = visitRepository.save(mapVisitModel(visitDto, offenderBooking))
+
+    addVisitors(visit, offenderBooking, visitDto)
```
andymarke, line 58: Probably worth looking at the mapping to understand why we need to do this as a separate step. Feels like we should be able to just create a Visit with a list of Visitors and save in one go rather then using a visitVistorRepository

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/resource/VisitResourceIntTest.kt
```diff
@@ -0,0 +1,114 @@
...
+        .body(BodyInserters.fromValue(createVisitRequest.copy(offenderNo = "Z9999ZZ")))
+        .exchange()
+        .expectStatus().isNotFound
+    }
+
+    @Test
+    fun `create visit success`() {
+      val response = webTestClient.post().uri("/visits")
+        .headers(setAuthorisation(roles = listOf("ROLE_UPDATE_NOMIS")))
+        .body(BodyInserters.fromValue(createVisitRequest))
```
andymarke, line 87: I think this would benefit if this was a inline String with the JSON rather than using CreateVisitRequest, given clients wouldn't use the CreateVisitRequest it helps to see the exact JSON being sent rather than relying on Jackson to convert an object to JSON.

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitServiceTest.kt
```diff
@@ -0,0 +1,181 @@
...
+      val balanceArgument = balanceArgumentCaptor.value
+      assertThat(balanceArgument.adjustReasonCode).isEqualTo(VisitOrderAdjustmentReason.VISIT_ORDER_ISSUE)
+      assertThat(balanceArgument.remainingVisitOrders).isEqualTo(-1)
+      assertThat(balanceArgument.remainingPrivilegedVisitOrders).isNull()
+      assertThat(balanceArgument.commentText).isEqualTo("Created by PVB3 for an on-line visit booking")
+
+      val visitArgumentCaptor = ArgumentCaptor.forClass(Visit::class.java)
+      verify(visitRepository).save(visitArgumentCaptor.capture())
+      val visit = visitArgumentCaptor.value
+      assertThat(visit.visitDate).isEqualTo(LocalDate.of(2021, 11, 4))
```
andymarke, line 134: LocalDate.parse("") tends to be more readable in tests the using `of`

Strings are more readable then numbers on the correct order

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitServiceTest.kt
```diff
@@ -0,0 +1,181 @@
...
+        )
+      )
+    )
+  }
+
+  @Nested
+  internal inner class createVisit {
+    @Test
+    fun success() {
+
```
andymarke, line 118: maybe break this test in to smaller ones since it is quite hard to follow what is important and why things happen?

fun `will return the visit id after creating`()

fun `will decrement the balance`()

fun `will set the core visit data`()

etc etc

andymarke (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitServiceTest.kt
```diff
@@ -0,0 +1,181 @@
...
+    @Test
+    fun successWithPrivilege() {
+
+      whenever(visitRepository.save(any())).thenReturn(Visit(id = visitId))
+
+      assertThat(visitService.createVisit(createVisitRequest.copy(privileged = true))).isEqualTo(CreateVisitResponse(visitId))
+
+      val balanceArgumentCaptor = ArgumentCaptor.forClass(OffenderVisitBalanceAdjustment::class.java)
+      verify(offenderVisitBalanceAdjustmentRepository).save(balanceArgumentCaptor.capture())
+      val balanceArgument = balanceArgumentCaptor.value
```
andymarke, line 162: you could probably use `check {}`  and assertj rather than an ArgumentCaptor

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/VisitVisitor.kt
steverendell: Ah good point, ill do that

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/resource/VisitResource.kt
steverendell: Ok, guess this makes sense as there could eventually be reference data endpoints or bookingId related ones

steverendell: Actually what would the amend/cancel urls be? This affects whether they can live in the same resource class

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/resource/VisitResource.kt
steverendell: Doesnt seem to - I can remove it

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitService.kt
steverendell: This looks tricky, im getting : 
Not-null property references a transient value - transient instance must be saved before current operation,
i.e. the Visitor's 'visit' property contains the raw transient Visit

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitService.kt
steverendell: id: Long = 0 🙂

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitService.kt
steverendell: done

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitService.kt
steverendell: done

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/resource/VisitResourceIntTest.kt
steverendell: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/resource/VisitResource.kt
steverendell: Done, for now they are PUTs using `/prisoner/offenderNo/{offenderNo}/visit/{visit}` and `/prisoner/offenderNo/{offenderNo}/visit/{visit}/cancel`

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitServiceTest.kt
steverendell: thanks, yes its much clearer (except I dont think I can use it in the inOrder case)

PR #17, opened 2022-01-19
SDI-15 Create visit nomis endpoint - sort event_id

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitService.kt
```diff
@@ -160,16 +162,27 @@ class VisitService(
...
+        eventId = getNextEvent()
+      )
     )
 
     visitDto.visitorPersonIds.forEach {
       val person = personRepository.findById(it).orElseThrow(DataNotFoundException("Person with id=$it does not exist"))
       visit.visitors.add(
-        VisitVisitor(visit = visit, person = person, eventStatus = scheduledEventStatus)
+        VisitVisitor(visit = visit, person = person, eventStatus = scheduledEventStatus, eventId = getNextEvent())
       )
```
andymarke, line 177: do you think you need it on each visitor as well as the dummy row?

andymarke (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitService.kt
```diff
@@ -160,16 +162,27 @@ class VisitService(
...
     }
   }
+
+  private fun getNextEvent(): Long {
+    // TODO Reviewing with Paul M as to whether this is a wise course to take!
+    // The stored proc uses event_id.nextval
+    return visitVisitorRepository.getEventId()
```
steverendell, line 184: Actually my current thinking is to just go with this so Ive not asked Paul yet

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitService.kt
steverendell: Ah good point - no, I think actually only needed on dummy row , will refine that next time

PR #19, opened 2022-01-24
SDI-28 write prefixed vsip visit id to nomis visit table

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Visit.kt
```diff
@@ -104,6 +104,9 @@ data class Visit(
   @JoinColumn(name = "AGENCY_VISIT_SLOT_ID")
   val agencyVisitSlot: AgencyVisitSlot? = null,
 
+  @Column(name = "CLIENT_UNIQUE_REF")
+  val vsipVisitId: String? = null,
+
```
lbennett-moj, line 109: I thought the name could reflect that we are commandeering this column

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/resource/VisitResourceIntTest.kt
```diff
@@ -134,6 +136,7 @@ class VisitResourceIntTest : IntegrationTestBase() {
         val visit = visitRepository.findById(response?.visitId!!).orElseThrow()
         assertThat(visit.endTime).isEqualTo(LocalDateTime.parse("2021-11-04T13:04"))
         assertThat(visit.offenderBooking?.bookingId).isEqualTo(offenderBookingId)
+        assertThat(visit.vsipVisitId).isEqualTo("VSIP_12345")
         assertThat(visit.visitors).extracting("person.id", "eventStatus.code").containsExactly(
```
lbennett-moj, line 140: prefix will be removed if returned in a dto in the future

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/data/CreateVisitRequest.kt
```diff
@@ -52,4 +52,8 @@ data class CreateVisitRequest(
   @Schema(description = "Location of visit, an agency internal location code", required = true)
   @NotBlank
   val visitRoomId: String,
+
+  @Schema(description = "VSIP visit id to allow mapping of nomis to VSIP visits", required = true)
+  @NotBlank
+  val vsipVisitId: String,
```
steverendell, line 58: Interesting one .. just thinking about whether this should be in the url: on the one hand a creation post rest call should probably not have an id in the url, but on the other it would be consistent with the cancel which will do

steverendell (approved):
lgtm with question

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/data/CreateVisitRequest.kt
lbennett-moj: hmmm, I think it works for the cancel but its part of the body to persist for the create, rather than an identifier

PR #20, opened 2022-01-25
SDI-24 Offender booking entity builder

andymarke:
Next steps 
1) Load PERSON (and contacts)
2) Change Repository tests to use builder
3) Add more tests around person and contacts

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
@@ -30,7 +30,7 @@ data class Offender(
   @GeneratedValue(generator = "OFFENDER_ID")
   @Id
   @Column(name = "OFFENDER_ID", nullable = false)
-  val id: Long,
+  var id: Long = 0,
 
   @Column(name = "OFFENDER_ID_DISPLAY", nullable = false)
   val nomsId: String,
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
```diff
@@ -30,7 +30,7 @@ data class Offender(
   @GeneratedValue(generator = "OFFENDER_ID")
   @Id
   @Column(name = "OFFENDER_ID", nullable = false)
-  val id: Long,
+  var id: Long = 0,
```
andymarke, line 34: assigned after creation

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderVisitBalance.kt
```diff
@@ -13,6 +16,15 @@ data class OffenderVisitBalance(
...
 
+  @OneToOne
+  @JoinColumn(
+    name = "OFFENDER_BOOK_ID",
+    referencedColumnName = "OFFENDER_BOOK_ID",
+    insertable = false,
+    updatable = false
+  )
+  val offenderBooking: OffenderBooking? = null,
+
```
andymarke, line 27: Feels like I should be able to combine this with the `@Id` field, but requires another PR to work that out

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderVisitBalance.kt
```diff
@@ -21,4 +33,19 @@ data class OffenderVisitBalance(
 
   @Column(name = "VISIT_ALLOWANCE_INDICATOR")
   val visitAllowanceIndicator: Boolean? = false,
-) : AuditableEntity()
+) {
+  override fun equals(other: Any?): Boolean {
```
andymarke, line 37: need else offenderBooking would be part of equals etc

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/resource/VisitResourceIntTest.kt
```diff
@@ -44,9 +49,32 @@ class VisitResourceIntTest : IntegrationTestBase() {
...
   @DisplayName("Create")
   @Nested
   inner class CreateVisitRequest {
+    lateinit var offenderNo: String
+    lateinit var offender: Offender
+    private var offenderBookingId: Long = 0L
+
+    @BeforeEach
+    internal fun createPrisoner() {
+      offender = repository.save(
```
andymarke, line 65: same offender is created for each test, we could do this per suite if it was too slow but obvious tests might interfere with each other (which in this case they won't)

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderVisitBalance.kt
steverendell: Im surpirsied it doesnt complain that the same col is mapped twice

steverendell (approved):
lgtm

PR #25, opened 2022-01-25
SDI-24 Improvement to OffenderVisitBalance mapping

andymarke:


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/Repository.kt
```diff
@@ -36,9 +30,9 @@ class Repository(
 
     offender.bookings.addAll(
       offenderBuilder.bookingBuilders.mapIndexed { index, bookingBuilder ->
-        val booking = save(bookingBuilder.build(offender, index, lookupAgency(bookingBuilder.agencyLocationId)))
```
andymarke, line 39: removed these unnecessary saves since it will get saved when the offender gets saved anyway

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderVisitBalance.kt
```diff
@@ -14,16 +15,12 @@ data class OffenderVisitBalance(
 
   @Id
   @Column(name = "OFFENDER_BOOK_ID", nullable = false)
-  val offenderBookingId: Long,
+  val offenderBookingId: Long = 0,
 
   @OneToOne
-  @JoinColumn(
```
andymarke, line 20: replace with MapsId

petergphillips (approved):
LGTM

PR #29, opened 2022-01-27
SDI-24 Use javaClass.hashCode() for any class that generates key so t…

andymarke:
…hat hashCode does not change during creation

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderVisitBalance.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderVisitBalance.kt
@@ -39,7 +39,7 @@ data class OffenderVisitBalance(
     return offenderBookingId == other.offenderBookingId
   }
 
-  override fun hashCode(): Int = offenderBookingId.hashCode()
+  override fun hashCode(): Int = javaClass.hashCode()
 
   @Override
   override fun toString(): String {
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderVisitBalanceAdjustment.kt
```diff
@@ -22,7 +21,11 @@ import javax.validation.constraints.Size
 @Entity
 @Table(name = "OFFENDER_VISIT_BALANCE_ADJS")
 data class OffenderVisitBalanceAdjustment(
-  @SequenceGenerator(name = "OFFENDER_VISIT_BALANCE_ADJ_ID", sequenceName = "OFFENDER_VISIT_BALANCE_ADJ_ID", allocationSize = 1)
+  @SequenceGenerator(
```
andymarke, line 24: just formatting

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Visit.kt
```diff
@@ -72,7 +71,13 @@ data class Visit(
           value = "'" + VisitStatus.VISIT_STATUS + "'",
           referencedColumnName = "domain"
         )
-      ), JoinColumnOrFormula(column = JoinColumn(name = "VISIT_STATUS", referencedColumnName = "code", nullable = false))
+      ), JoinColumnOrFormula(
```
andymarke, line 74: just formatting

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderVisitBalance.kt
```diff
@@ -39,7 +39,7 @@ data class OffenderVisitBalance(
     return offenderBookingId == other.offenderBookingId
   }
 
-  override fun hashCode(): Int = offenderBookingId.hashCode()
+  override fun hashCode(): Int = javaClass.hashCode()
```
andymarke, line 43: interestingly this key not generated but use the bookingId which itself is generated, so this can change during creation

petergphillips (approved):
LGTM

PR #37, opened 2022-02-08
SDI-23 Cancel visit nomis endpoint - tidying

steverendell:


build.gradle.kts
```diff
@@ -33,7 +33,6 @@ dependencies {
 
   testImplementation("org.awaitility:awaitility-kotlin:4.1.1")
   testImplementation("io.jsonwebtoken:jjwt:0.9.1")
-  testImplementation("com.github.tomakehurst:wiremock-standalone:2.27.2")
```
steverendell, line 36: not needed

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitService.kt
```diff
@@ -68,11 +68,10 @@ class VisitService(
     val offenderBooking = offenderBookingRepository.findByOffenderNomsIdAndActive(offenderNo, true)
       .orElseThrow(NotFoundException(offenderNo))
 
-    val existingVisits = visitRepository.findByVsipVisitId(VSIP_PREFIX + visitDto.vsipVisitId)
-    if (!existingVisits.isEmpty()) {
-      val existingVisit = existingVisits[0]
+    val existingVisit = visitRepository.findOneByVsipVisitId(VSIP_PREFIX + visitDto.vsipVisitId)
+    if (existingVisit.isPresent) {
```
lbennett-moj, line 72: it might be nice to have a repository function returning a kotlin nullable to avoid having to work with optional

lbennett-moj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitService.kt
steverendell: good point, will do

steverendell: actaully that bit is getting removed along with the VSIP id!

PR #71, opened 2022-05-16
SDI-192 Add phones numbers to lead visitor

andymarke:
* Had to some entity attributes up to constructor so we can create phone and addresses
* Builders can now create a person with addresses and phone numbers
* Visitor endpoint now returns the lead visitor (if exists) with name and phone numbers

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/data/VisitResponse.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/data/VisitResponse.kt
@@ -2,6 +2,7 @@ package uk.gov.justice.digital.hmpps.nomisprisonerapi.data
 
 import com.fasterxml.jackson.annotation.JsonInclude
 import io.swagger.v3.oas.annotations.media.Schema
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.Phone
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.Visit
 import java.time.LocalDateTime
 import javax.validation.constraints.NotEmpty
@@ -126,9 +127,19 @@ data class VisitResponse(
     visitorConcernText = visitEntity.visitorConcernText,
     visitors = visitEntity.visitors.filter { visitor -> visitor.person != null }
       .map { visitor -> Visitor(visitor.person!!.id, visitor.groupLeader) },
-    visitOutcome = null,
-    leadVisitor = null
+    visitOutcome = null, // TODO
+    leadVisitor = visitEntity.visitors.find { visitor -> visitor.groupLeader }?.person?.let {
+      LeadVisitor(
+        personId = it.id,
+        fullName = "${it.firstName} ${it.lastName}",
+        telephones = it.phones.toTelephoneList() + it.addresses.flatMap { address -> address.phones.toTelephoneList() }
+      )
+    }
   )
 
   data class CodeDescription(val code: String, val description: String)
+}
+
+fun List<Phone>.toTelephoneList(): List<String> {
+  return this.map { phone -> phone.phoneNo ?: "" }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Address.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Address.kt
@@ -26,7 +26,18 @@ import javax.persistence.Table
 @Table(name = "ADDRESSES")
 @DiscriminatorColumn(name = "OWNER_CLASS")
 @Inheritance
-abstract class Address {
+abstract class Address(
+  val premise: String? = null,
+  val street: String? = null,
+  val locality: String? = null,
+  @Column(name = "START_DATE")
+  val startDate: LocalDate = LocalDate.now(),
+  @Column(name = "NO_FIXED_ADDRESS_FLAG")
+  val noFixedAddressFlag: String = "N",
+  @OneToMany(mappedBy = "address", cascade = [CascadeType.ALL], fetch = FetchType.LAZY, orphanRemoval = true)
+  @Where(clause = "OWNER_CLASS = '" + AddressPhone.PHONE_TYPE + "'")
+  val phones: MutableList<AddressPhone> = ArrayList()
+) {
   @Id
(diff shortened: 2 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Address.kt
```diff
@@ -26,7 +26,18 @@ import javax.persistence.Table
...
+  @Column(name = "START_DATE")
+  val startDate: LocalDate = LocalDate.now(),
+  @Column(name = "NO_FIXED_ADDRESS_FLAG")
+  val noFixedAddressFlag: String = "N",
+  @OneToMany(mappedBy = "address", cascade = [CascadeType.ALL], fetch = FetchType.LAZY, orphanRemoval = true)
+  @Where(clause = "OWNER_CLASS = '" + AddressPhone.PHONE_TYPE + "'")
+  val phones: MutableList<AddressPhone> = ArrayList()
+) {
   @Id
   @SequenceGenerator(name = "ADDRESS_ID", sequenceName = "ADDRESS_ID", allocationSize = 1)
```
andymarke, line 42: moved attributes that we now use to create up to constructor

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AddressPhone.kt
```diff
@@ -8,19 +8,15 @@ import javax.persistence.ManyToOne
 
 @Entity
 @DiscriminatorValue(AddressPhone.PHONE_TYPE)
-data class AddressPhone(
+class AddressPhone(
```
andymarke, line 11: no advantage of it being a data class given its super class isn't

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/PersonAddressBuilder.kt
```diff
@@ -0,0 +1,31 @@
...
+
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.AddressPhone
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.Person
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.PersonAddress
+
+class PersonAddressBuilder(
+  var premise: String = "41",
+  var street: String = "High Street",
+  var locality: String = "Sheffield",
+  private var phoneNumbers: List<Pair<String, String>> = listOf()
```
andymarke, line 11: no bothered to do much more than these 3, a future version could set city and country if required.

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/data/VisitResponse.kt
```diff
@@ -126,9 +127,19 @@ data class VisitResponse(
...
+        telephones = it.phones.toTelephoneList() + it.addresses.flatMap { address -> address.phones.toTelephoneList() }
+      )
+    }
   )
 
   data class CodeDescription(val code: String, val description: String)
 }
+
+fun List<Phone>.toTelephoneList(): List<String> {
+  return this.map { phone -> phone.phoneNo ?: "" }
```
petergphillips, line 144: this looks a bit odd to me, wouldn't it be better to do a `mapNonNull` rather than end up with empty phone numbers in the list?

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Address.kt
```diff
@@ -26,7 +26,18 @@ import javax.persistence.Table
...
+abstract class Address(
+  val premise: String? = null,
+  val street: String? = null,
+  val locality: String? = null,
+  @Column(name = "START_DATE")
+  val startDate: LocalDate = LocalDate.now(),
+  @Column(name = "NO_FIXED_ADDRESS_FLAG")
+  val noFixedAddressFlag: String = "N",
+  @OneToMany(mappedBy = "address", cascade = [CascadeType.ALL], fetch = FetchType.LAZY, orphanRemoval = true)
+  @Where(clause = "OWNER_CLASS = '" + AddressPhone.PHONE_TYPE + "'")
```
petergphillips, line 38: would be better to do string interpolation here rather than adding 3 strings together

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/PersonAddress.kt
```diff
@@ -1,18 +1,32 @@
...
-  val person: Person
-) : Address() {
+  @ManyToOne(optional = false, fetch = LAZY)
+  val person: Person,
+  premise: String? = null,
+  street: String? = null,
+  locality: String? = null,
+  startDate: LocalDate = LocalDate.now(),
+  noFixedAddressFlag: String = "N",
+  phones: MutableList<AddressPhone> = ArrayList()
```
petergphillips, line 21: shouldn't this be `mutableListOf` as well?

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/PersonAddress.kt
```diff
@@ -1,18 +1,32 @@
...
 import javax.persistence.DiscriminatorValue
 import javax.persistence.Entity
-import javax.persistence.FetchType
+import javax.persistence.FetchType.LAZY
 import javax.persistence.JoinColumn
 import javax.persistence.ManyToOne
 
 @Entity
 @DiscriminatorValue(PersonAddress.ADDR_TYPE)
-data class PersonAddress(
```
petergphillips, line 11: why is this no longer a `data` class?

petergphillips (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/PersonAddress.kt
andymarke: @petergphillips I think I will keep it as a plain class; this inherits from a plain class so we just don't have any of the benefits of a data class; there is only one attribute on the old data class - and making it a data class forces you to add an ugly constructor to allow for the parameters that need to be passed to the super class.

PR #73, opened 2022-05-17
SDI-192 Order numbers by last updated and add extensions

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Phone.kt
```diff
@@ -16,9 +17,9 @@ import javax.persistence.Table
 @Inheritance
 abstract class Phone(
   @Column(name = "PHONE_TYPE")
-  val phoneType: String? = null,
+  val phoneType: String,
```
andymarke, line 20: non-null on database

lbennett-moj (approved):
LGTM

PR #76, opened 2022-05-18
SDI-192 Returned outcome code even if there is reference data link

andymarke:
There are 3 outcome reason codes that do not exist in the NOMIS reference data table. This allows the code to still be returned to the called with a dummy description


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/data/VisitResponse.kt
```diff
@@ -127,8 +128,9 @@ data class VisitResponse(
     visitorConcernText = visitEntity.visitorConcernText,
     visitors = visitEntity.visitors.filter { visitor -> visitor.person != null }
       .map { visitor -> Visitor(visitor.person!!.id, visitor.groupLeader) },
-    visitOutcome = visitEntity.visitors.find { it.person == null }?.outcomeReason
-      ?.let { CodeDescription(it.code, it.description) },
+    visitOutcome = visitEntity.outcomeVisitor()?.outcomeReason
+      ?.let { CodeDescription(it.code, it.description) }
+      ?: visitEntity.outcomeVisitor()?.outcomeReasonCode?.let { CodeDescription(it, it) },
     leadVisitor = visitEntity.visitors.find { visitor -> visitor.groupLeader }?.person?.let {
```
andymarke, line 134: if we have no ref data return code (and description as code)

lbennett-moj (approved):

PR #81, opened 2022-06-07
SDI-220 Removed update of audit fields, allow trigger to populate them

andymarke:
When OFFENDER_VISIT_BALANCE_ADJS was updated the CREATE_USER_ID was being set to client creds. The preference is to allow the trigger to set it to the database username instead to be consistent with all other tables.


petergphillips (approved):
LGTM

PR #97, opened 2022-07-19
SDI-273: 🔧 Dual run both ingress controllers

petergphillips:


petergphillips:
not relevant as not running in cloud platform

PR #101, opened 2022-08-02
SDI-305 New algorithm for generating NOMIS description for a given room

andymarke:
Description needs to follows NOMIS idioms while being able to describe the room and open closed state.

`
"Main visit room" for open visits become HEI-VSIP-MAIN-SOC 
"Main visit room" for closed visits become HEI-VSIP-MAIN-CLO
`


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/VisitServiceTest.kt
```diff
@@ -152,45 +150,6 @@ internal class VisitServiceTest {
     visitors.add(VisitVisitor(visit = this, person = Person(-7L, "First", "Last")))
   }
 
-  private val defaultLocation = AgencyLocation(id = "SWI", description = "Swindon")
```
andymarke, line 155: don't need these, since Mocks changed to echo back what was supplied for the rep `save` method

steverendell (approved):
lgtm

PR #106, opened 2022-08-31
SDI-355 add current flag to get incentive and add new get current inc…

lbennett-moj:
…entive endpoint


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/IncentivesService.kt
```diff
@@ -35,19 +35,35 @@ class IncentivesService(
         offenderBooking = offenderBooking,
         sequence = incentiveSequence
       )
-    )?.let { mapIncentiveModel(it) }
+    )?.let {
+      // determine if this incentive is the current IEP for the booking
+      val currentIep =
+        incentiveRepository.findFirstById_offenderBookingOrderByIepDateDescId_SequenceDesc(offenderBooking)
+          ?.let { current -> current == it } ?: false
```
steverendell, line 42: i *think possibly this is equivalent: 
val currentIep = it ==
        incentiveRepository.findFirstById_offenderBookingOrderByIepDateDescId_SequenceDesc(offenderBooking)

steverendell (dismissed):
lgtm with a comment

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/IncentivesService.kt
lbennett-moj: yes - thanks, makes it simpler

steverendell (approved):
lgtm

PR #111, opened 2022-09-12
SDI-360 Update Nomis from incentive creation domain event

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderBooking.kt
```diff
@@ -113,6 +113,15 @@ data class OffenderBooking(
   @OneToMany(mappedBy = "id.offenderBooking", cascade = [CascadeType.ALL], fetch = FetchType.LAZY)
   val incentives: MutableList<Incentive> = mutableListOf(),
 ) {
+  fun getNextSequence(): Long {
+    val i = incentives.stream().max(Comparator.comparing { (a, _) -> a.sequence })
```
andymarke, line 117: You could have used Kotlin semantics here `incentives.map(it.sequence).maxOrNull()?. let {it + 1}   ?: 1`  or something like that

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/IncentivesService.kt
```diff
@@ -1,27 +1,64 @@
...
+  companion object {
+    private val log = LoggerFactory.getLogger(this::class.java)
+  }
+
+  fun createIncentive(bookingId: Long, dto: CreateIncentiveRequest): CreateIncentiveResponse {
+    val offenderBooking = offenderBookingRepository.findById(bookingId)
+      .orElseThrow(NotFoundException(bookingId.toString()))
+
+    val mappedIncentive = mapIncentiveModel(dto, offenderBooking)
+    val incentive = incentiveRepository.save(mappedIncentive)
```
andymarke, line 45: not sure why we need to save here? Can't we just add to booking and let the transaction commit?

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/IncentivesService.kt
```diff
@@ -67,4 +104,25 @@ class IncentivesService(
...
+  private fun mapIncentiveModel(dto: CreateIncentiveRequest, offenderBooking: OffenderBooking): Incentive {
+
+    val sequence = offenderBooking.getNextSequence()
+
+    val iepLevel = iepLevelRepository.findById(IEPLevel.pk(dto.iepLevel))
+      .orElseThrow(BadDataException("Invalid IEP type from incentives: ${dto.iepLevel}"))
+
+    val location = agencyLocationRepository.findById(dto.agencyId)
+      .orElseThrow(BadDataException("Prison with id=${dto.agencyId} does not exist"))
+
```
andymarke, line 117: I think you also need to check the incentive level exist for the prison as well (sorry can't remember table)
In fact, if you just check IEP level/location exists then the 2 above calls are redundant I think?

andymarke (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderBooking.kt
steverendell: Im glad you mentioned that ! , i couldnt work out what the 'native' way forward was with this

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/service/IncentivesService.kt
steverendell: good point

andymarke (approved):
LGTM

PR #113, opened 2022-09-12
SDI-376 room usage should filter out historical visits

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/specification/VisitSpecification.kt
```diff
@@ -42,8 +43,8 @@ class VisitSpecification(private val filter: VisitFilter) : Specification<Visit>
       predicates.add(criteriaBuilder.lessThanOrEqualTo(root.get(Visit::whenCreated.name), this))
     }
 
-    filter.ignoreMissingRoom?.takeIf { it }?.run {
-      predicates.add(criteriaBuilder.isNotNull(root.get<String>(Visit::agencyInternalLocation.name)))
+    filter.futureVisits?.takeIf { it }?.run {
+      predicates.add(criteriaBuilder.greaterThan(root.get(Visit::visitDate.name), LocalDate.now()))
     }
```
andymarke, line 48: maybe it should be greater or equal then ?

andymarke (approved):
LGTM

PR #116, opened 2022-09-20
SDI-385 Add update visit

andymarke:
* Allows updates to the visitor list

TODO all other mutable fields


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/data/UpdateVisitRequest.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude
+import io.swagger.v3.oas.annotations.media.Schema
+import java.time.LocalDateTime
+import java.time.LocalTime
+import javax.validation.constraints.NotEmpty
+import javax.validation.constraints.NotNull
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@Schema(description = "Visit update request")
+data class UpdateVisitRequest(
```
andymarke, line 12: assumed prisonId never changes so not included

PR #117, opened 2022-09-21
SDI-385 Add update visit

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/data/UpdateVisitRequest.kt
```diff
@@ -23,10 +23,6 @@ data class UpdateVisitRequest(
   @NotEmpty
   val visitorPersonIds: List<Long>,
 
-  @Schema(description = "Visit type, whether social or official", allowableValues = ["SCON", "OFFI"], required = true)
-  @NotEmpty
-  val visitType: String,
-
```
andymarke, line 29: no need support changing visit from social to official right now, VSIP doesn't allow this

mikehalmamoj (approved):

PR #126, opened 2022-10-03
SDI-360 Update Nomis from incentive creation domain event

steverendell:
Fix to IEP_TIME


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Incentive.kt
```diff
@@ -39,7 +39,7 @@ data class Incentive(
   val iepDate: LocalDate,
 
   @Column(nullable = false)
-  val iepTime: LocalTime,
+  val iepTime: LocalDateTime,
```
steverendell, line 42: In Nomis this has an important date part

lbennett-moj (commented):
LGTM

andymarke (approved):
Lgtm

PR #129, opened 2022-10-12
SDI-417 expose audit fields for visits

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Visit.kt
```diff
@@ -120,6 +120,12 @@ data class Visit(
   @Column(name = "CREATE_DATETIME", nullable = false, insertable = false, updatable = false)
   var whenCreated: LocalDateTime = LocalDateTime.now(),
 
+  @Column(name = "MODIFY_DATETIME", nullable = false, insertable = false, updatable = false)
+  var whenUpdated: LocalDateTime? = null,
+
+  @Column(name = "CREATE_USER_ID", insertable = false, updatable = false)
+  val createUserId: String = "",
+
```
lbennett-moj, line 128: never updated as these fields are handled by oracle. Just avoids having to include the field in the constructor, not sure if there is a better way

mikehalmamoj (approved):

PR #145, opened 2022-12-21
SDI-517: ✨ Upgrade to spring boot 3

petergphillips:

PR #156, opened 2023-01-13
SDI-536 Skeleton of sentence adjustment resource

andymarke:
Added basic mapping for sentence adjustment.
This requires to create tables for all related entities when running locally/tests due to foreign key constraints even though none of them are mapped


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderSentence.kt
```diff
@@ -0,0 +1,38 @@
...
+
+@Embeddable
+data class SentenceId(
+  @ManyToOne(optional = false, fetch = FetchType.LAZY)
+  @JoinColumn(name = "OFFENDER_BOOK_ID", nullable = false)
+  val offenderBooking: OffenderBooking,
+
+  @Column(name = "SENTENCE_SEQ", nullable = false)
+  val sequence: Long
+) : Serializable
```
andymarke, line 22: Not interested in any other column right now, when just need the primary key

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderSentenceAdjustment.kt
```diff
@@ -0,0 +1,88 @@
...
+  @Column(name = "SENTENCE_SEQ", nullable = false)
+  val sequence: Int,
+
+  @ManyToOne(optional = false, fetch = FetchType.LAZY)
+  @JoinColumns(
+    value = [
+      JoinColumn(
+        name = "OFFENDER_BOOK_ID",
+        referencedColumnName = "OFFENDER_BOOK_ID",
+        insertable = false,
```
andymarke, line 63: maybe we can get rid of the double mapping of this foreign keys - not sure yet

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/SentenceAdjustment.kt
```diff
@@ -0,0 +1,16 @@
...
+
+@Entity
+@Table(name = "SENTENCE_ADJUSTMENTS")
+class SentenceAdjustment(
+  @Id
+  @Column(name = "SENTENCE_ADJUST_CODE", nullable = false)
+  val id: String,
+  @Column(name = "DESCRIPTION", nullable = false)
+  val description: String,
+)
```
andymarke, line 16: only map the two columns we are interested in, will end up creating reference data in the flyway next

src/main/resources/db/migration/V1_12__SENTENCE_CALC_TYPES.sql
```diff
@@ -0,0 +1,56 @@
+CREATE TABLE "SENTENCE_CALC_TYPES"
```
andymarke, line 1: all these copied from prison-api

src/main/resources/db/migration/V2_1_SENTENCE_ADJUSTMENTS.sql
```diff
@@ -0,0 +1,38 @@
+create table SENTENCE_ADJUSTMENTS
```
andymarke, line 1: this not copied from prison-api so added v2 so the numbers never clash as we copy more files

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderSentence.kt
```diff
@@ -0,0 +1,38 @@
...
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.ManyToOne
+import jakarta.persistence.Table
+import org.hibernate.Hibernate
+import java.io.Serializable
+
+@Embeddable
+data class SentenceId(
+  @ManyToOne(optional = false, fetch = FetchType.LAZY)
+  @JoinColumn(name = "OFFENDER_BOOK_ID", nullable = false)
```
mikehalmamoj, line 17: Do you need nullable = true for when we move to Hibernate 6? (remember the issue Peter found) https://github.com/ministryofjustice/hmpps-nomis-prisoner-api/pull/145#pullrequestreview-1225939691

andymarke: I just don't know - I suspect not since the problem we had was for a foreign key where actually other entity didn't exist, in other words there was no foreign key constraint and we had bad data

src/main/resources/db/migration/V1_12__SENTENCE_CALC_TYPES.sql
mikehalmamoj: Are we sure prison-api is up to date?

mikehalmamoj (approved):

PR #157, opened 2023-01-16
SDI-536 Basic sentence mapping so we can create test data

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderSentence.kt
```diff
@@ -23,9 +29,27 @@ data class SentenceId(
 
 @Entity
 @Table(name = "OFFENDER_SENTENCES")
+// Warning: on mandatory fields mapped - just enough to create test entities
 data class OffenderSentence(
```
andymarke, line 33: no doubt we will fill in the rest some time later this year when we will actually create sentences in production

lbennett-moj (approved):
LGTM

PR #177, opened 2023-01-31
SDI-500 validate amend activity pay band

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/PayBand.kt
```diff
@@ -0,0 +1,14 @@
...
+
+@Entity
+@DiscriminatorValue(PayBand.PAY_BAND)
+class PayBand(code: String, description: String) : ReferenceCode(PAY_BAND, code, description) {
+
+  companion object {
+    const val PAY_BAND = "PAY_BAND"
+    fun pk(code: String): Pk = Pk(PAY_BAND, code)
+  }
+}
```
mikehalmamoj, line 14: So we need to add a repo and some seed data for the pay bands.

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderProgramProfilePayBand.kt
```diff
@@ -22,10 +25,24 @@ data class OffenderProgramProfilePayBand(
   @Column(nullable = false)
   val startDate: LocalDate,
 
+  @Id
```
steverendell, line 28: Dont think this should be an id column? PK is oPP + start date

steverendell (approved):
lgtm

PR #181, opened 2023-02-01
ll SDI-549 Update to sentence adjustment

andymarke:
- Assume all data will be resent
- Assume booking and sentence relationship can not change


petergphillips (approved):
LGTM

PR #182, opened 2023-02-01
SDI-500 Prevent deletion of pay rate if allocated to an offender

mikehalmamoj:
This validation occurs in Nomis so we should do the same in our Nomis update API.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/ActivitiesService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/ActivitiesService.kt
@@ -299,6 +299,17 @@ class ActivitiesService(
       .filter { old -> !old.hasFutureStartDate() } // ignore future rates not included in updates - so they are deleted
       .filter { old -> !newPayRates.containsRate(old) }
       .map { old -> old.expire() }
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/ActivitiesService.kt
```diff
@@ -299,6 +299,17 @@ class ActivitiesService(
...
+      .also { expiredPayRates -> expiredPayRates.throwIfPayBandsInUse() }
+
+  private fun List<CourseActivityPayRate>.throwIfPayBandsInUse() =
+    this.forEach { activityPayRate ->
+      offenderProgramProfileRepository.findByCourseActivity(activityPayRate.id.courseActivity)
+        .filter { profile -> profile.isUsingPayBand(activityPayRate.payBand.code) }
+        .map { offender -> offender.offenderBooking.offender.nomsId }
+        .toList()
+        .takeIf { nomsIds -> nomsIds.isNotEmpty() }
+        ?.run { throw BadDataException("Pay band ${activityPayRate.payBand.code} is allocated to offender(s) $this") }
```
andymarke, line 311: Feels like this can be shorter, not sure why you need the map for instance. Feels like it can be a "find" or something

andymarke (approved):
Lgtm

PR #188, opened 2023-02-02
SDI-500 Limit the update request to what we know will need updating

mikehalmamoj:
But the details of what else should change and what should happen on a change need defining (e.g. does an end date end all of the pay rates and deallocate all prisoner from that date?) and will be covered on SDI-599


andymarke (approved):
lgtm

PR #200, opened 2023-02-09
SDI-613 Split up the Activity service and its tests

mikehalmamoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/Repository.kt
```diff
@@ -186,6 +186,7 @@ class Repository(
     agencyInternalLocationRepository.findById(locationId).orElse(null)
 
   fun delete(offender: Offender) = offenderRepository.deleteById(offender.id)
+  fun deleteOffenders() = offenderRepository.deleteAll()
```
steverendell, line 189: just need to check at some point that any 'static' seed data that gets deleted isnt needed

steverendell (approved):
lgtm

PR #203, opened 2023-02-10
SDI-613 refactor pay rates tests to unit tests

mikehalmamoj:


src/main/resources/db/migration/data/R__1_15__OFFENDERS.sql
```diff
@@ -1,2 +0,0 @@
-INSERT INTO OFFENDERS (OFFENDER_ID, ID_SOURCE_CODE, LAST_NAME, MIDDLE_NAME, FIRST_NAME, SEX_CODE, CREATE_DATE, LAST_NAME_KEY, OFFENDER_ID_DISPLAY, ROOT_OFFENDER_ID, RACE_CODE, ALIAS_NAME_TYPE, BIRTH_DATE, BIRTH_COUNTRY_CODE)
```
mikehalmamoj, line 1: Removed a load of transactional data that's not used in any tests

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/CourseActivityBuilder.kt
```diff
@@ -54,19 +58,57 @@ class CourseActivityBuilder(
   var payRates: List<CourseActivityPayRateBuilder>,
 ) {
   fun build(): CourseActivity =
+    repository?.let {
+      CourseActivity(
+        code = code,
+        program = repository.lookupProgramService(programId),
```
mikehalmamoj, line 64: If the repository has been injected this must be an integration test so load related entities.

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/CourseActivityBuilder.kt
```diff
@@ -54,19 +58,57 @@ class CourseActivityBuilder(
...
+      ?: throw IllegalStateException("No repository - is this a unit test? Try create() instead.")
+
+  fun create(
+    programCode: String = "CA",
+    internalLocationCode: String? = "CRM1"
+  ): CourseActivity =
     CourseActivity(
       code = code,
-      program = repository.lookupProgramService(programId),
+      program = programService(programCode),
```
mikehalmamoj, line 86: In unit tests we call the `create` function instead which instantiates related entities manually (as there is no database).

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/PayRatesService.kt
```diff
@@ -89,7 +89,7 @@ class PayRatesService(
     }
 
   private fun CourseActivityPayRate.rateIsUnchanged(requested: PayRateRequest) =
-    this.halfDayRate.compareTo(requested.rate) == 0
+    this.halfDayRate.compareTo(CourseActivityPayRate.preciseHalfDayRate(requested.rate)) == 0
```
steverendell, line 92: Why not use .equals() ?

steverendell (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/PayRatesService.kt
mikehalmamoj: Because that's what [Kotlin does under the hood](https://kotlinlang.org/docs/equality.html)

mikehalmamoj: Oh I see you mean why compareTo instead of equals! I think it was because equals doesn't work with BigDecimals

PR #213, opened 2023-02-20
SDI-635 Create Nomis endpoint to create appointment in Nomis

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/appointments/AppointmentsResource.kt
```diff
@@ -0,0 +1,69 @@
...
+import org.springframework.web.bind.annotation.ResponseStatus
+import org.springframework.web.bind.annotation.RestController
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.config.ErrorResponse
+
+@RestController
+@Validated
+@RequestMapping(produces = [MediaType.APPLICATION_JSON_VALUE])
+class AppointmentsResource(private val appointmentService: AppointmentService) {
+  @PreAuthorize("hasRole('ROLE_NOMIS_APPOINTMENTS')")
+  @PostMapping("/appointments")
```
andymarke, line 24: Not sure it matters too ,much but do you think this should be 
`/prisoners/booking-id/{bookingId}/appointments`  so it is consistent with IEPs and sentencing adjustments?

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/appointments/AppointmentService.kt
```diff
@@ -0,0 +1,72 @@
...
+          "appointment-created",
+          mapOf(
+            "id" to it.id.toString(),
+            "bookingId" to it.offenderBooking.bookingId.toString(),
+            "location" to it.internalLocation?.locationId.toString(),
+          ),
+          null
+        )
+      }
+      .let { CreateAppointmentResponse(offenderIndividualScheduleRepository.save(it).id) }
```
andymarke, line 40: Not sure it matters too much, but should we model this as a booking having many appointments, so this code would end up looking like 
`booking.addAppointment(mapModel(...))` so no need for the repository?

andymarke: The disadvantage you will need to flush to the get the newly created id that is a bit ugly, see sentencing adjustments

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderIndividualSchedule.kt
```diff
@@ -0,0 +1,108 @@
...
+
+  @Enumerated(EnumType.STRING)
+  @Column(nullable = false)
+  val eventClass: EventClass = EventClass.INT_MOV, // INT_MOV only for now
+
+  @Column(nullable = false)
+  val eventType: String = "APP", // APP for appointment
+
+  @ManyToOne(optional = false)
+  @NotFound(action = NotFoundAction.EXCEPTION)
```
andymarke, line 58: maybe have a look at preprod and see if we ever have missing values in which case you could drop the NotFound

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderIndividualSchedule.kt
```diff
@@ -0,0 +1,108 @@
...
+          referencedColumnName = "domain"
+        )
+      ), JoinColumnOrFormula(column = JoinColumn(name = "EVENT_STATUS", referencedColumnName = "code"))
+    ]
+  )
+  val eventStatus: EventStatus,
+
+  @ManyToOne(fetch = FetchType.LAZY)
+  @JoinColumn(name = "AGY_LOC_ID")
+  val prison: AgencyLocation? = null,
```
andymarke, line 87: so I guess both these columns are nullable? Interesting to know why, I can imaging why internal location is null but weird that AGY_LOC_ID is null. BTW is it always a prison or can it be something else lie a court or whatever?

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/appointments/AppointmentsResourceIntTest.kt
```diff
@@ -0,0 +1,236 @@
...
+    offenderAtOtherPrison = repository.save(
+      OffenderBuilder(nomsId = "A1234TU")
+        .withBooking(OffenderBookingBuilder(agencyLocationId = "LEI"))
+    )
+  }
+
+  @AfterEach
+  internal fun deletePrisoner() {
+    repository.delete(offenderAtMoorlands)
+  }
```
andymarke, line 48: looks like you have forgotten to delete `offenderAtOtherPrison`

andymarke (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/appointments/AppointmentsResource.kt
steverendell: Actually I think appointments are conceptually a bit different because they are transitory - i.e. the event happens, then it expires. So the concept of a prisoner having a 'list' of appointments is not really important.

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/appointments/AppointmentService.kt
steverendell: As above

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderIndividualSchedule.kt
steverendell: 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderIndividualSchedule.kt
steverendell: from the data, TAPs have a null AGY_LOC_ID

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderIndividualSchedule.kt
steverendell: Changed to ignore as there could be existing data with codes not in the ref list 🤷

andymarke (approved):
lgtm

PR #220, opened 2023-02-21
SDI-556 A flag for the hidden sentence adjustments

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderSentenceAdjustment.kt
```diff
@@ -55,6 +55,9 @@ class OffenderSentenceAdjustment(
   @Column(name = "SENTENCE_SEQ", nullable = false)
   val sentenceSequence: Long,
```
andymarke, line 57: We don't need to map this to an actual object since we only need to know the existence of the foreign key

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/SentenceBuilder.kt
```diff
@@ -44,6 +44,7 @@ class SentenceAdjustmentBuilder(
   var adjustmentDate: LocalDate = LocalDate.now(),
   var createdDate: LocalDateTime = LocalDateTime.now(),
   var adjustmentNumberOfDays: Long = 10,
+  var kayDateAdjustmentId: Long? = null,
```
mikehalmamoj, line 47: typo

lbennett-moj (approved):

PR #235, opened 2023-03-08
SDI-651 Capture telemetry when schedule rules updated

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/ActivityService.kt
```diff
@@ -106,12 +108,16 @@ class ActivityService(
       existingActivity.courseScheduleRules.addAll(newRules)
     }
 
+    activityRepository.saveAndFlush(existingActivity)
+
```
mikehalmamoj, line 112: I gave in and now save the entity before the @Transactional annotation because I need to know the ids of any saved schedule rules so I can use them in telemetry.

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/PayRatesService.kt
```diff
@@ -158,34 +153,28 @@ class PayRatesService(
     )
   }
 
-  private fun publishTelemetry(
-    courseActivityId: Long,
```
mikehalmamoj, line 162: This is now published along with its parent so we just build the telemetry in this service

petergphillips (approved):
LGTM

PR #243, opened 2023-04-04
SDI-689 Derive the schedule instance

mikehalmamoj:


andymarke (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/AttendanceService.kt
```diff
@@ -34,29 +36,38 @@ class AttendanceService(
     createAttendanceRequest: CreateAttendanceRequest,
   ): CreateAttendanceResponse =
     attendanceRepository.save(createAttendanceRequest.toOffenderCourseAttendance(scheduleId, bookingId))
-      .let { CreateAttendanceResponse(it.eventId) }
+      .let { CreateAttendanceResponse(it.eventId, it.courseSchedule!!.courseScheduleId) }
 
-  private fun CreateAttendanceRequest.toOffenderCourseAttendance(scheduleId: Long, bookingId: Long): OffenderCourseAttendance {
-    val courseSchedule = scheduleRepository.findByIdOrNull(scheduleId)
-      ?: run { throw NotFoundException("Course schedule with id=$scheduleId not found") }
+  private fun CreateAttendanceRequest.toOffenderCourseAttendance(courseActivityId: Long, bookingId: Long): OffenderCourseAttendance {
```
andymarke, line 41: maybe this would read better not being an extension function, and just pass in the request given you need the repo looks up from the enclosing class?

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/AttendanceService.kt
```diff
@@ -34,29 +36,38 @@ class AttendanceService(
...
+    val courseActivity = activityRepository.findByIdOrNull(courseActivityId)
+      ?: throw NotFoundException("Course activity with id=$courseActivityId not found")
+
+    val courseSchedules = scheduleRepository.findByCourseActivityAndScheduleDate(courseActivity, scheduleDate)
```
andymarke, line 45: could you add a `findByCourseActivityAndScheduleDateAndStartTimeAndEndTime` and get rid of the filter, and also make it return a single optional rather than a list ( and hence get rid of the switch below)?

andymarke (approved):
LGTM

PR #249, opened 2023-04-11
SDIT-690 add telemetry for upsert attendance

mikehalmamoj:
And add validation to prevent updates if already paid in Nomis


andymarke (approved):
lgtm

PR #258, opened 2023-04-14
SDIT-688 Add missing column

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderCourseAttendance.kt
```diff
@@ -133,6 +133,9 @@ class OffenderCourseAttendance(
 
   @Column(name = "REFERENCE_ID")
   val referenceId: Long? = null,
+
+  @Column(name = "PERFORMANCE_CODE")
+  val performanceCode: String? = "STANDARD",
```
mikehalmamoj, line 138: No captured by Activities so default in the same way as is done in Whereabouts

lbennett-moj (approved):
LGTM

PR #262, opened 2023-04-18
SDIT-743 Make API consistent

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/api/UpdateCourseScheduleRequest.kt
```diff
@@ -10,7 +10,7 @@ import java.time.LocalTime
 data class UpdateCourseScheduleRequest(
 
   @Schema(description = "The date of the course schedule", example = "2023-04-03", required = true)
-  val scheduleDate: LocalDate,
+  val date: LocalDate,
```
mikehalmamoj, line 13: Other schedule requests use `date` instead of `scheduleDate` which caused a client to get it wrong.

petergphillips (approved):
LGTM

PR #278, opened 2023-04-26
SDIT-412 Enhance Update Activity API

mikehalmamoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/ActivityResourceIntTest.kt
```diff
@@ -469,7 +482,7 @@ class ActivityResourceIntTest : IntegrationTestBase() {
 
         callUpdateEndpoint(
           courseActivityId = existingActivity.courseActivityId,
-          jsonBody = updateActivityRequestJson(internalLocationJson = """"internalLocationId: -99999,"""),
+          jsonBody = updateActivityRequestJson(detailsJson = detailsJson().replace(""""internalLocationId" : -27,""", """"internalLocationId: -99999,""")),
```
steverendell, line 485: I think we need a better way of modifying template test json. Maybe use data classes with an unmarshaller rather than literal json

mikehalmamoj: Hmmm, unmarhsalling some JSON just to marshal it with the same json library is convenient but can hide bugs IMHO. I prefer to throw raw JSON at the endpoint as it's closer to the real world.

mikehalmamoj: It also means I can test that the types on my API are correct by sending invalid types?

mikehalmamoj: I could drop the `replace` and go back to the JSON builder method I had before - I just found the builder incredibly hard to read but if it made the tests easier to read then would that be better?

steverendell: Maybe one to discuss next Thurs

PR #292, opened 2023-05-18
SDIT-816 Return ID of specific created adjustment rather than last one

andymarke:
When the create endpoint is called at the exact same time it is possible that one of the calls returns the ID of the wrong newly created adjustment, this ensures it always returns the ID of the one just created


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingAdjustmentService.kt
```diff
@@ -54,7 +54,7 @@ class SentencingAdjustmentService(
   fun createSentenceAdjustment(bookingId: Long, sentenceSequence: Long, request: CreateSentenceAdjustmentRequest) =
     offenderBookingRepository.findByIdOrNull(bookingId)?.let { booking ->
       offenderSentenceRepository.findByIdOrNull(SentenceId(booking, sentenceSequence))?.let { sentence ->
-        sentence.adjustments.add(
+        val adjustmentId = offenderSentenceAdjustmentRepository.save(
```
andymarke, line 57: save adjustment directly rather than adding to list and saving parent, we can't guarantee that `adjustments.last()` below will return this new one or another created at exact same time

mikehalmamoj (approved):

PR #293, opened 2023-05-19
SDIT-817 Added tables required adjudications

andymarke:
First draft of adding tables to H2 required to support adjudications.

Schemas taken from live but file names taken from prison-api.
Where prison-api has the order wrong and cheated by disabling constraints added to the 1000 range


Mjwillis:
Lgtm

PR #308, opened 2023-06-13
SDIT-838 Include course schedule ids update on requests

mikehalmamoj:
But ignore them for now - need to pass them in the calling service before we can start using


petergphillips (approved):
LGTM

PR #309, opened 2023-06-14
SDIT-838 Use ids to match existing schedules

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/ScheduleService.kt
```diff
@@ -66,28 +66,49 @@ class ScheduleService(
...
+    if (scheduleDate <= LocalDate.now() &&
+      (
+        scheduleDate != requested.scheduleDate ||
+          startTime != requested.startTime ||
+          endTime != requested.endTime
+        )
+    ) {
+      throw BadDataException("Cannot update schedules starting before tomorrow")
+    }
+  }
```
mikehalmamoj, line 98: Retain this restriction for now to limit the size of this change - now we don't rely on the business key we can change it whenever, coming soon.

PR #311, opened 2023-06-15
Sdit 870 create adj

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/IncidentDecisionAction.kt
```diff
@@ -0,0 +1,16 @@
...
+import jakarta.persistence.DiscriminatorValue
+import jakarta.persistence.Entity
+
+@Entity
+@DiscriminatorValue(IncidentDecisionAction.INC_DECISION)
+class IncidentDecisionAction(code: String, description: String) : ReferenceCode(INC_DECISION, code, description) {
+
+  companion object {
+    const val INC_DECISION = "INC_DECISION"
+    const val NO_FURTHER_ACTION_CODE = "NFA"
```
andymarke, line 12: I suspect we can remove the above const - I bet we never have cause to use it

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/Repository.kt
```diff
@@ -293,6 +315,20 @@ class Repository(
   fun lookupProgramEndReason(code: String): ProgramServiceEndReason =
     programEndReasonRepository.findByIdOrNull(ProgramServiceEndReason.pk(code))!!
 
+  fun lookupStaff(id: Long): Staff =
```
andymarke, line 318: remove it

andymarke (approved):
LGTM

PR #318, opened 2023-06-19
sdit_870 more wiring up of jpa to response dtos and H2 trigger

lbennett-moj:


andymarke (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AdjudicationIncidentParty.kt
```diff
@@ -87,4 +87,4 @@ class AdjudicationIncidentParty(
   override fun hashCode(): Int = javaClass.hashCode()
 }
 
-fun AdjudicationIncidentParty.asSuspect(): OffenderBooking = this.offenderBooking.takeIf { this.incidentRole == suspectRole }!!
+fun AdjudicationIncidentParty.prisonerOnReport(): OffenderBooking = offenderBooking.takeIf { this.adjudicationNumber != null }!!
```
mikehalmamoj, line 90: Is this to force an exception if the adjudication number is null?

PR #331, opened 2023-06-23
SDIT-856 Parties test

andymarke:


lbennett-moj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AdjudicationIncidentParty.kt
```diff
@@ -18,6 +18,11 @@ import java.io.Serializable
 import java.time.LocalDate
 
 const val suspectRole = "S"
+const val witnessRole = "W"
+const val victimRole = "V"
+const val reportingOfficerRole = "RO"
+const val forceControllingOfficerRole = "CR"
+const val otherRole = "OTH"
```
mikehalmamoj, line 25: Should these come from the PartyRole enum?

andymarke: Actually the other way around, the enum uses these values. Enum is for DSL and these are need because we need to understand the values in the business logic.

BTW it is one of the few instances were we don't map to a ReferenceCode class because the column maps to 2 different domains depending on context

PR #334, opened 2023-06-23
SDIT-843 Allow update attendance in wrong prison or allocation ended

mikehalmamoj:
But prevent these if creating an attendance

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/AttendanceService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/AttendanceService.kt
@@ -74,7 +74,14 @@ class AttendanceService(
         offenderBooking,
         offenderProgramProfile,
         status,
-      )
+      ).also {
+        if (offenderBooking.location?.id != courseSchedule.courseActivity.prison.id) {
+          throw BadDataException("Prisoner is at prison=${offenderBooking.location?.id}, not the Course activity prison=${courseSchedule.courseActivity.prison.id}")
+        }
+        if (offenderProgramProfile.programStatus.code == "END") {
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/AttendanceService.kt
```diff
@@ -143,11 +150,8 @@ class AttendanceService(
     offenderBooking: OffenderBooking,
     bookingId: Long,
   ) =
-    offenderProgramProfileRepository.findByCourseActivityCourseActivityIdAndOffenderBookingBookingIdAndProgramStatusCode(
-      courseSchedule.courseActivity.courseActivityId,
-      offenderBooking.bookingId,
-      "ALLOC",
-    )
+    offenderProgramProfileRepository.findByCourseActivityAndOffenderBooking(courseSchedule.courseActivity, offenderBooking)
+      .maxByOrNull { it.startDate }
```
mikehalmamoj, line 154: As we can udpate if allocation ended don't restrict to status ALLOC - just get the latest

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderProgramProfile.kt
```diff
@@ -39,7 +39,7 @@ data class OffenderProgramProfile(
   var program: ProgramService,
 
   @Column(name = "OFFENDER_START_DATE")
-  val startDate: LocalDate? = null,
```
mikehalmamoj, line 42: This column cannot be null even though table column is nullable. So point in wasting time catering for null values.

andymarke: and I guess you never read this column and need to read old bad NOMIS data?

andymarke (approved):
lgtm

PR #337, opened 2023-06-26
SDIT-902 more course activity test data DSLs

mikehalmamoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/ActivityResourceIntTest.kt
```diff
@@ -681,24 +676,26 @@ class ActivityResourceIntTest : IntegrationTestBase() {
...
-          courseActivity = courseActivity {
-            payRate(iepLevelCode = "STD")
-            payRate(iepLevelCode = "BAS")
+          programService {
+            courseActivity = courseActivity {
+              payRate(iepLevelCode = "STD")
+              payRate(iepLevelCode = "BAS")
+              courseSchedule()
+              courseScheduleRule()
+            }
```
mikehalmamoj, line 698: This is the point of the PR - filled out the entity hierarchy from program service -> course activity -> children. 
One thing I'm not sure of is that when using builders only we used to default a single pay rate, course schedule and course schedule rule which could all be overridden with an empty list or a list full of stuff as the test dictated. Now we default an empty list due to the DSL pattern we need to specify those single defaults - can't decide whether to go back to defauling them again or keeping their declaration explicit 🤔

PR #338, opened 2023-06-27
SDIT-907 Allow creation of Nomis attendance from outside prison.

mikehalmamoj:
E.g. when suspended after transfer, and possibly due to other Activities business rules.


andymarke (approved):
lgtm

PR #347, opened 2023-06-29
SDIT-911 Don't overwrite a confirmed attendance

mikehalmamoj:
There is a screen in Nomis called Internal Movements which can be used to confirm an internal movement (such as an attendance) and sets the event status to COMP. This is separate to marking of the attendance and shouldn't be overridden.


petergphillips (approved):
LGTM

PR #352, opened 2023-07-03
SDIT-870 adjudication hearing result award mappings and DSL

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AdjudicationHearingResultAward.kt
```diff
@@ -0,0 +1,154 @@
...
+        nullable = false,
+      ),
+      JoinColumn(
+        name = "RESULT_SEQ",
+        referencedColumnName = "RESULT_SEQ",
+        nullable = false,
+      ),
+    ],
+  )
+  val hearingResult: AdjudicationHearingResult?,
```
lbennett-moj, line 68: one exists in production without an associated hearing

PR #354, opened 2023-07-04
SDIT-871 Reorganise test repository

andymarke:
- Adjudication DSL interface now contains all defaults, with defaults moved from Builder constructor
- Repository reorganised into sections
- Helpers used in Tests (rather than in Builders) moved and renamed to use the convention of `get` rather than `lookup` to disiguish from Builder reference data lookups


petergphillips (approved):
LGTM

PR #355, opened 2023-07-05
SDIT-916 Handle course start date moved back in time

mikehalmamoj:
Handles an edge case where the course start date is movedbefore the existing pay rates become active by adjusting the start date of the pay rates.


andymarke (approved):
lgtm

PR #356, opened 2023-07-05
SDIT-916 Fix further edge cases for changing course start date

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/PayRatesService.kt
```diff
@@ -175,6 +181,7 @@ class PayRatesService(
             .toCourseActivityPayRate(
               courseActivity = existingActivity,
               startDate = existingActivity.scheduleStartDate,
+              endDate = it.endDate,
```
mikehalmamoj, line 184: When creating the new pay rate with an earlier start date bear in mind it is not necessarily active - it could have an end date.

petergphillips (approved):
LGTM

PR #358, opened 2023-07-06
SDIT-926 GET /adjudications/ids resource

andymarke:
* Can filter a list of adjudications by fromDate, toDate and a list of prisons


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationService.kt
```diff
@@ -84,7 +85,8 @@ class AdjudicationService(
     pageRequest: Pageable,
     adjudicationFilter: AdjudicationFilter,
   ): Page<AdjudicationIdResponse> {
-    return Page.empty()
+    return adjudicationIncidentPartyRepository.findAll(AdjudicationSpecification(adjudicationFilter), pageRequest)
+      .map { AdjudicationIdResponse(adjudicationNumber = it.adjudicationNumber!!, offenderNo = it.offenderBooking!!.offender.nomsId) }
   }
```
andymarke, line 90: we can use !! since we only find parties that have adjudication numbers and they must be prisoners (rather than staff)

mikehalmamoj (approved):

PR #365, opened 2023-07-12
SDIT-873 Refactor Adjudication builders

andymarke:
- Move DSL interface definitions to domain specific files
- Remove DSL sub-interfaces and fold into main interface since parent entity builder already has knowledge of the interface since it has to implement it
- Rename DSL file so it represents the top level data builder


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/NomisData.kt
```diff
@@ -0,0 +1,143 @@
+package uk.gov.justice.digital.hmpps.nomisprisonerapi.helper.builders
```
andymarke, line 1: actually a rename of DSL.kt

mikehalmamoj (approved):

PR #367, opened 2023-07-14
SDIT-873 Move sentence adjustments to use DSL

andymarke:


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/AdjustmentRepositoryTest.kt
```diff
@@ -46,9 +46,9 @@ class AdjustmentRepositoryTest {
     builderRepository.save(
       LegacyOffenderBuilder()
         .withBooking(
```
andymarke, line 48: can not use DSL here until we figure out class loading dependencies

lbennett-moj (approved):

PR #391, opened 2023-08-01
SDIT-969 Set created user and fix issue with prisoners with aliases

andymarke:
* Now sets created user on Incident since this is displayed in NOMIS
* Fix issues where root offender was not being retrieved so prisoners with aliases had "no bookings"

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderRepository.kt
@@ -1,11 +1,16 @@
 package uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.repository
 
 import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Query
 import org.springframework.stereotype.Repository
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.Offender
 
 @Repository
 interface OffenderRepository : JpaRepository<Offender, Long> {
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderRepository.kt
```diff
@@ -1,11 +1,16 @@
...
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.Offender
 
 @Repository
 interface OffenderRepository : JpaRepository<Offender, Long> {
   fun findByNomsId(nomsId: String): List<Offender>
-  fun findFirstByNomsId(nomsId: String): Offender?
+
+  @Query("select o from Offender o left join fetch o.bookings b WHERE o.nomsId = :nomsId order by b.bookingSequence asc")
+  fun findByNomsIdOrderedWithBookings(nomsId: String): List<Offender>
 }
```
andymarke, line 14: copied from prison-api since this has proved to be reliable on selecting the correct record

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderRepository.kt
```diff
@@ -1,11 +1,16 @@
...
+
+fun OffenderRepository.findRootByNomisId(nomsId: String): Offender? = findByNomsIdOrderedWithBookings(nomsId).firstOrNull()
```
andymarke, line 16: convince method to emphasises what it is retrieving

mikehalmamoj (approved):

PR #394, opened 2023-08-01
SDIT-980 find migration activities - paging

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/CourseActivityRepository.kt
```diff
@@ -1,8 +1,24 @@
...
+    select ca.courseActivityId from CourseActivity ca 
+    where ca.prison.id = :prisonId and ca.courseActivityId in 
+      (
+       select distinct(opp.courseActivity.courseActivityId) from OffenderProgramProfile opp
+       where opp.prison.id = :prisonId and opp.programStatus.code = 'ALLOC'
+      )   
+  """,
+  )
+  fun findActivitiesToMigrate(prisonId: String, pageRequest: Pageable): Page<Long>
+}
```
mikehalmamoj, line 24: Next up - adding more rules to the query

mikehalmamoj: I hate the syntax of specifications. I can read and understand SQL. So I'm going to try SQL first and switch if I really have to.

PR #406, opened 2023-08-07
SDIT-1025 Don't allow attendance create after an allocation has ended

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/AttendanceService.kt
```diff
@@ -75,8 +75,10 @@ class AttendanceService(
         offenderProgramProfile,
         requestStatus,
       ).also {
-        if (offenderProgramProfile.programStatus.code == "END") {
-          throw BadDataException("Cannot create an attendance for allocation ${offenderProgramProfile.offenderProgramReferenceId} because it has ended")
```
mikehalmamoj, line 79: Status of END isn't good enough - it's the date it ended that is of more relevance.

steverendell (approved):
lgtm

PR #407, opened 2023-08-08
SDIT-981 Include pay per session

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/api/GetActivityResponse.kt
```diff
@@ -46,6 +46,9 @@ data class GetActivityResponse(
   @Schema(description = "Whether the course runs on bank holidays", example = "false")
   val excludeBankHolidays: Boolean,
 
+  @Schema(description = "Half or Full day (H or F)", required = true, example = "H")
```
karenmillermoj, line 49: Very minor - if this is the response, does it need "required = true" ?

mikehalmamoj: No! Copy and paste laziness. I'll fix in another PR.

PR #409, opened 2023-08-08
SDIT-981 Exclude programs when finding ids

mikehalmamoj:
We'll be synchronising activities back to Nomis with new program codes once they are migrated - we'll need to exclude these from being migrated themselves.


steverendell (approved):
lgtm

steverendell (approved):
lgtm

PR #411, opened 2023-08-10
SDIT-927 Additional adjudication data required by DPS

andymarke:
DPS requires
* username for all staff
* evidence and repairs require the user who created these records in NOMIS
* Incident created time
* Gender of prisoner
* Current establishment of prisoner
Also fix null adjudicationNumber that cannot be null when retrieving by adjudication number

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AdjudicationEvidence.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AdjudicationEvidence.kt
@@ -56,6 +56,9 @@ class AdjudicationEvidence(
   @ManyToOne(fetch = FetchType.LAZY)
   @JoinColumn(name = "AGY_INC_INVESTIGATION_ID")
   val investigation: AdjudicationInvestigation,
+
+  @Column(name = "CREATE_USER_ID", insertable = false, updatable = false)
+  val createUsername: String = "",
 ) {
 
   override fun equals(other: Any?): Boolean {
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AdjudicationEvidence.kt
```diff
@@ -56,6 +56,9 @@ class AdjudicationEvidence(
   @ManyToOne(fetch = FetchType.LAZY)
   @JoinColumn(name = "AGY_INC_INVESTIGATION_ID")
   val investigation: AdjudicationInvestigation,
+
+  @Column(name = "CREATE_USER_ID", insertable = false, updatable = false)
+  val createUsername: String = "",
 ) {
```
andymarke, line 62: only ever read so just default to empty string

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AdjudicationEvidence.kt
```diff
@@ -56,6 +56,9 @@ class AdjudicationEvidence(
   @ManyToOne(fetch = FetchType.LAZY)
   @JoinColumn(name = "AGY_INC_INVESTIGATION_ID")
   val investigation: AdjudicationInvestigation,
+
+  @Column(name = "CREATE_USER_ID", insertable = false, updatable = false)
+  val createUsername: String = "",
```
mikehalmamoj, line 61: lateinit var?

mikehalmamoj (approved):

PR #412, opened 2023-08-14
SDIT-927 Additional adjudication data required by DPS

andymarke:
DPS requires additional data for migration

- User who created the witnesses and victims
- User who created a hearing result
- Date time the result was created
- The NOMIS sanction sequence

In addition only outcomes for the request charge are returned for multi-charge adjudications

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationService.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.nomisprisonerapi.adjudications
 
+import jakarta.persistence.EntityManager
 import org.springframework.data.domain.Page
 import org.springframework.data.domain.Pageable
 import org.springframework.data.repository.findByIdOrNull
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationService.kt
```diff
@@ -214,7 +216,11 @@ class AdjudicationService(
           createRepairForAdjudicationIncident(incident = this, index + 1, repair)
         }
       }
-      .let { mapAdjudication(it.parties.findAdjudication(adjudicationNumber)) }
+      .let {
+        // created user needs refreshing from database - hence the flush and clear
+        adjudicationIncidentRepository.flush().also { entityManager.clear() }
+      }
```
andymarke, line 222: jump through hops so that the entity is re-read from database since the username is set by database/trigger not by JPA

mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationService.kt
mikehalmamoj: Why isn't this a flush and refresh instead of clear? Also wondering if you could use Hibernate's [Generated](https://docs.jboss.org/hibernate/orm/6.2/javadocs/org/hibernate/annotations/Generated.html) annotation instead?

PR #423, opened 2023-08-23
SDIT-1043 Api to end activities and allocations

mikehalmamoj:
To be used after migrating Nomis activities to DPS.


petergphillips (approved):
LGTM

PR #430, opened 2023-08-29
SDIT-1043 Add optional end activity / allocation comment

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/AllocationService.kt
```diff
@@ -236,9 +236,10 @@ class AllocationService(
     agencyLocationRepository.findByIdOrNull(prisonId)
       ?: throw BadDataException("Prison with id=$prisonId does not exist")
 
-  fun endAllocation(courseAllocation: OffenderProgramProfile, date: LocalDate) {
+  fun endAllocation(courseAllocation: OffenderProgramProfile, date: LocalDate, endAllocationComment: String? = null) {
     courseAllocation.endDate = date
     courseAllocation.endReason = findEndReasonOrThrow("OTH")
     courseAllocation.programStatus = findProgramStatus("END")
+    endAllocationComment?.run { courseAllocation.endComment = endAllocationComment }
   }
```
andymarke, line 244: curious, why can't you just do `courseAllocation.endComment = endAllocationComment`   is it because it would over write something when null?

andymarke (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/AllocationService.kt
mikehalmamoj: It could overwrite an existing end reason - but I'm still not sure why I would pass a null end comment (just didn't want to default the words "Migrated" because that's very specific for a general endpoint).

But it is odd - his needs more thought.

PR #437, opened 2023-08-30
SDIT-874 Return the create username and set module

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationService.kt
```diff
@@ -436,6 +440,7 @@ class AdjudicationService(
       )
     }
     adjudicationParty.incident.repairs.addAll(updatedRepairs)
+    adjudicationIncidentPartyRepository.saveAndFlush(adjudicationParty)
```
andymarke, line 443: force entity reload so create username is set correctly

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AdjudicationIncidentRepairs.kt
```diff
@@ -53,10 +54,11 @@ class AdjudicationIncidentRepair(
...
   val type: AdjudicationRepairType,
-
-  @Column(name = "CREATE_USER_ID", insertable = false, updatable = false)
-  val createUsername: String = "",
 ) {
+  @Column(name = "CREATE_USER_ID", insertable = false, updatable = false)
+  @Generated
+  lateinit var createUsername: String
+
   override fun equals(other: Any?): Boolean {
```
andymarke, line 62: now consistent with other entities

lbennett-moj (approved):

PR #441, opened 2023-09-01
SDIT-1072 Switch to DTO HQL query for IDs, so no entity loads

andymarke:
Using the specification had safe effect of loading child entities when all we need is 3 fields


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationResource.kt
```diff
@@ -68,7 +67,7 @@ class AdjudicationResource(
     ],
   )
   fun getAdjudicationChargeIdsByFilter(
-    @PageableDefault(sort = ["id.agencyIncidentId", "id.chargeSequence"], direction = Sort.Direction.ASC, size = 20)
```
andymarke, line 71: now hard coded in HQL

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/AdjudicationIncidentChargeRepository.kt
```diff
@@ -1,12 +1,42 @@
...
+    """
+      select 
+        charge.incidentParty.adjudicationNumber as adjudicationNumber, 
+        charge.id.chargeSequence as chargeSequence, 
+        charge.incidentParty.offenderBooking.offender.nomsId as nomsId
+      from AdjudicationIncidentCharge charge 
+        where 
+          (:fromDate is null or charge.whenCreated > :fromDate) and 
+          (:toDate is null or charge.whenCreated < :toDate) and   
+          (:hasPrisonFilter = false or charge.incident.prison.id in :prisonIds)
```
andymarke, line 25: extra dummy flag required since no way of checking if prisonIds is null in HQL

mikehalmamoj (approved):

PR #446, opened 2023-09-05
SDIT-1072 Force locations as open so proxy works

andymarke:
Entities are not open by default, nor can we make all entities open it causes too many compile errors, so just make 2 of them that are unnecessary loading LAZY child bart creating custom annotation


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AdjudicationIncidentParty.kt
```diff
@@ -53,7 +53,6 @@ class AdjudicationIncidentPartyId(
       attributeNodes = [
         NamedAttributeNode("offender", subgraph = "offender-with-gender"),
         NamedAttributeNode("location"),
-        NamedAttributeNode("createLocation"),
```
andymarke, line 56: no longer needed in join since the proxy now works

petergphillips (approved):
LGTM

PR #450, opened 2023-09-07
Sdit 1084 adjudication additional data

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationResponse.kt
```diff
@@ -60,23 +60,7 @@ data class AdjudicationCharge(
   val reportDetail: String?,
   val offenceId: String?,
   val chargeSequence: Int,
-) {
```
andymarke, line 63: found a better way of doing this, e.g filter out award before converting to DTO

lbennett-moj (approved):
lgtm

PR #455, opened 2023-09-08
SDIT-1080 Amend non-associaton

steverendell:
add sequence to url


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/nonassociations/NonAssociationService.kt
```diff
@@ -136,45 +149,42 @@ class NonAssociationService(
       throw BadDataException("Effective date must not be in the future")
     }
 
-    existing.nonAssociationReason = recipReason
-    existing.recipNonAssociationReason = recipReason
+    existing.nonAssociation.nonAssociationReason = recipReason
+    existing.nonAssociation.recipNonAssociationReason = recipReason
```
karenmillermoj, line 153: I know you haven't changed the value being set - but just checking they have the same value?

karenmillermoj (approved):
lgtm

PR #459, opened 2023-09-15
SDIT-997 Add updatedBy to non-associations response

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/nonassociations/NonAssociationService.kt
```diff
@@ -348,7 +348,13 @@ class NonAssociationService(
...
           authorisedBy = detail.authorisedBy,
+          updatedBy = detail.updatedBy,
           comment = detail.comment,
         )
       }
 }
+
+private val OffenderNonAssociationDetail.updatedBy: String
+  get() {
+    return modifiedBy ?: createdBy!!
```
karenmillermoj, line 359: annoyingly have to add !! as createdBy is set to nullable

steverendell: Just checked visits and there the field is not nullable, but initialised to be blank - maybe try that?

steverendell (approved):

steverendell (approved):
lgtm

PR #463, opened 2023-09-20
SDIT-996 Add sync service from Nomis

steverendell:
Add Audit for NAs to generate DPS_SYNCHRONISATION value


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/nonassociations/NonAssociationService.kt
```diff
@@ -34,6 +35,7 @@ class NonAssociationService(
   private val typeRepository: ReferenceCodeRepository<NonAssociationType>,
   private val telemetryClient: TelemetryClient,
 ) {
+  @Audit
```
karenmillermoj, line 38: I'm assuming we don't test anything for this?

karenmillermoj (approved):
lgtm

PR #464, opened 2023-09-20
SDIT-983 Use defaults where possible for Open API docs

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/ActivitiesResource.kt
```diff
@@ -310,9 +297,6 @@ class ActivitiesResource(
       ApiResponse(
         responseCode = "200",
         description = "OK",
-        content = [
-          Content(mediaType = "application/json", schema = Schema(implementation = FindActiveActivityIdsResponse::class)),
```
mikehalmamoj, line 314: Unnecessary, and in this case also inaccurate so it was breaking generated code in a client.

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/ActivitiesResource.kt
```diff
@@ -141,7 +134,7 @@ class ActivitiesResource(
     ],
   )
   fun updateActivity(
-    @Schema(description = "Course activity id", required = true) @PathVariable courseActivityId: Long,
```
mikehalmamoj, line 144: Unnecessary, let the required status be derived.

lbennett-moj (approved):
lgtm

PR #473, opened 2023-09-28
SDIT-1096 Non-association Reconciliation process

steverendell:
Order NA prisoner pairs by id, not nomsId


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderNonAssociationRepository.kt
```diff
@@ -20,7 +20,7 @@ interface OffenderNonAssociationRepository :
          from OffenderNonAssociation na, Offender o1, Offender o2
```
karenmillermoj, line 20: Comment refers to nomsId but we now use offenderId

karenmillermoj (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderNonAssociationRepository.kt
steverendell: whoops yes, will fix next time

PR #478, opened 2023-10-04
SDIT-1101 New endpoint to end multiple activities and their allocations

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderProgramProfileRepository.kt
```diff
@@ -51,4 +53,18 @@ interface OffenderProgramProfileRepository : JpaRepository<OffenderProgramProfil
...
+  @Query(
+    nativeQuery = true,
+    value = """
+      update offender_program_profiles opp
+      set opp.offender_end_date = :date, opp.offender_end_reason = 'OTH', opp.offender_program_status = 'END'
+      where opp.crs_acty_id in :courseActivityIds
+      and opp.offender_program_status = 'ALLOC'
+      and (opp.offender_end_date is null or opp.offender_end_date > current_date)
+    """,
+  )
```
mikehalmamoj, line 68: Had to use a native query here as the reference codes being updated caused some very strange SQL when trying to use JPQL - probably something to do with the join formulae on the ManyToOne relationship.

steverendell: Is this for 'old' activities that are no longer needed? Presumably some of the migrated activities will continue to be used?

steverendell (approved):
lgtm with minor Q

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderProgramProfileRepository.kt
mikehalmamoj: No ,we migrate the old activities to DPS which are then synchronised back to NOMIS on new program services. At this point we have duplicate activities in NOMIS - so this endpoint will be used to end all of the old activities.

Which I should probably have mentioned in the PR desciption 🤦

PR #484, opened 2023-10-11
SDIT-883 Create award returns the sanction sequence

andymarke:
Add values to swagger so that we have an enum on the client generated code


steverendell (approved):
lgtm

PR #487, opened 2023-10-16
SDIT-1151 referral outcomes need to update existing outcome

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationService.kt
```diff
@@ -614,28 +614,53 @@ class AdjudicationService(
...
+    val existingResult = adjudicationHearingResultRepository.findFirstOrNullById_OicHearingIdAndChargeSequence(
+      chargeSequence = chargeSequence,
+      hearingId = hearingId,
+    )
 
+    existingResult?.let {
+      it.pleaFindingType = lookupPleaFindingType(request.pleaFindingCode)
+      it.pleaFindingCode = request.pleaFindingCode
+      it.findingType = lookupFindingType(request.findingCode)
+      adjudicationHearingResultRepository.saveAndFlush(it)
```
lbennett-moj, line 653: save and flushing here to stop an app insights created message when DB write could fail. Maybe we don't usually bother but trying to avoid confusion on error

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationService.kt
```diff
@@ -644,39 +669,18 @@ class AdjudicationService(
...
-              "chargeSequence" to chargeSequence.toString(),
-              "hearingId" to hearingId.toString(),
-              "resultSequence" to "1",
-            ),
+            telemetryMap,
             null,
           )
         }
-    } else {
-      telemetryClient.trackEvent(
```
lbennett-moj, line 661: don't need the ignore now and update service is mapping "REF_INAD" finding codes

andymarke (approved):
LGTM

PR #489, opened 2023-10-17
SDIT-1138 Set consecutive award

andymarke:
Given an adjudication this will find the first award that is the same type of award that is being added. In practice this will always be an ADA (unless DPS rules change)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AdjudicationHearingResultAward.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AdjudicationHearingResultAward.kt
@@ -72,8 +72,6 @@ class AdjudicationHearingResultAward(
     name = "OIC_INCIDENT_ID",
     referencedColumnName = "OIC_INCIDENT_ID",
     nullable = false,
-    insertable = false,
-    updatable = false,
   )
   val incidentParty: AdjudicationIncidentParty,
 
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AdjudicationHearingResultAward.kt
```diff
@@ -72,8 +72,6 @@ class AdjudicationHearingResultAward(
     name = "OIC_INCIDENT_ID",
     referencedColumnName = "OIC_INCIDENT_ID",
     nullable = false,
-    insertable = false,
-    updatable = false,
   )
```
andymarke, line 75: this was accidentally never being set for awards

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationsHearingResultAwardResourceIntTest.kt
```diff
@@ -229,7 +305,87 @@ class AdjudicationsHearingResultAwardResourceIntTest : IntegrationTestBase() {
             assertThat(it).containsEntry("hearingId", existingHearing.id.toString())
             assertThat(it).containsEntry("adjudicationNumber", existingAdjudicationNumber.toString())
             assertThat(it).containsEntry("resultSequence", "1")
-            assertThat(it).containsEntry("sanctionSequence", "1")
+            assertThat(it).containsEntry("sanctionSequence", "4")
```
andymarke, line 308: sequence now 4 since we have 3 awards on a previous adjudication

mikehalmamoj (approved):

PR #490, opened 2023-10-17
SDIT-1152 Return an error code when trying to update a paid attendance

mikehalmamoj:
So that the client can react to this specific error case


andymarke (approved):
lgtm

PR #493, opened 2023-10-19
SDIT-1165 KEY_DATE now KEY-DATE so matches with mapping service

andymarke:
This value gets passed all the way through the migration service to the mapping service; but that is expecting KEY-DATE


ldlharper (approved):
LGTM!

PR #500, opened 2023-10-24
SDIT-1161 update award now deletes removed awards

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationService.kt
```diff
@@ -727,36 +727,41 @@ class AdjudicationService(
     adjudicationNumber: Long,
     chargeSequence: Int,
     requests: CreateHearingResultAwardRequest,
-  ): CreateHearingResultAwardResponses =
-    CreateHearingResultAwardResponses(
+  ): CreateHearingResultAwardResponses {
+    val party = adjudicationIncidentPartyRepository.findByAdjudicationNumber(adjudicationNumber)
```
andymarke, line 731: moved these out so sanctionSeq is calculated before any deletes happen

PR #502, opened 2023-10-27
SDIT-1136 Attendance reconciliation endpoint

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderCourseAttendanceRepository.kt
```diff
@@ -1,15 +1,32 @@
...
+      from OffenderCourseAttendance oca
+      join OffenderBooking ob on oca.offenderBooking = ob
+      where oca.prison.id = :prisonId
+      and oca.eventDate = :date
+      and oca.pay = true
+      group by oca.offenderBooking.bookingId
+      order by oca.offenderBooking.bookingId
+    """,
+  )
+  fun findBookingPaidAttendanceCountsByPrisonAndDate(prisonId: String, date: LocalDate): List<BookingCount>
```
mikehalmamoj, line 31: This is the main change - get all bookings and attendances with pay_flag=Y so that we can compare with the data in DPS to report any differences.

petergphillips (approved):
LGTM

PR #505, opened 2023-10-31
SDIT-884 delete punishments

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationService.kt
```diff
@@ -837,10 +837,26 @@ class AdjudicationService(
     return UpdateHearingResultAwardResponses(awardsCreated = createdAwards, awardsDeleted = deletedAwards)
   }
 
-  private fun deletedRemovedHearingResultAwards(
+  @Audit
```
andymarke, line 840: Very very minor, but redundant since we are only deleting so there is no audit column updated

andymarke (approved):
Lgtm

PR #506, opened 2023-10-31
SDIT-793 Add agency feature switches to /info endpoint

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/health/FeatureSwitchInfoContributor.kt
```diff
@@ -0,0 +1,21 @@
...
+import org.springframework.stereotype.Component
+import org.springframework.transaction.annotation.Transactional
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.core.ServiceAgencySwitchesService
+
+@Component
+class FeatureSwitchInfoContributor(
+  private val serviceAgencySwitchesService: ServiceAgencySwitchesService,
+) : InfoContributor {
+
+  @Transactional(readOnly = true)
```
petergphillips, line 14: this looks a bit odd - especially as it is calling a different service anyway.  Would have expected the `readonly = true` to be on that service instead?  Perhaps default at the top of the file should be `readonly` and then changed as needed for methods that amend data?

petergphillips (approved):
LGTM

PR #517, opened 2023-11-10
SDIT-1234 Don't break the API contract until clients switched over

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/api/UpsertAllocationRequest.kt
```diff
@@ -50,7 +50,7 @@ data class UpsertAllocationRequest(
   val programStatusCode: String,
 
   @Schema(description = "Sessions excluded from the allocation during which period attendances will not be generated")
-  val exclusions: List<AllocationExclusion> = listOf(),
+  val exclusions: List<AllocationExclusion>? = listOf(),
```
mikehalmamoj, line 53: Make the new request backwards compatible so that the client doesn't have to be released in lockstep. Can make this mandatory again once the client has been updated and deployed.

andymarke (approved):
lgtm

PR #524, opened 2023-11-13
SDIT-1222 Return error code if prisoner deallocated and moved

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/data/Exceptions.kt
```diff
@@ -10,6 +10,7 @@ class BadDataException(message: String?, val error: BadRequestError? = null) : R
 
 enum class BadRequestError(val errorCode: Int) {
   ATTENDANCE_PAID(1001),
+  PRISONER_MOVED_ALLOCATION_ENDED(1002),
```
mikehalmamoj, line 13: We need to catch this specific error in the client so attach an error code to it.

steverendell (approved):
lgtm

PR #529, opened 2023-11-16
SDIT-1226 reference data

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offence.kt
```diff
@@ -0,0 +1,34 @@
...
+@Embeddable
+class OffenceId(
+  val offenceCode: String,
+  val statuteCode: String,
+) : Serializable
+
+@Entity
+@Table(name = "OFFENCES")
+@EntityOpen
+class Offence(
```
mikehalmamoj, line 20: You might want to add a toString if not data class

andymarke (approved):
LGTM

PR #541, opened 2023-11-30
SDIT-1226 court order, mappings, builders etc

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/CourtOrder.kt
```diff
@@ -0,0 +1,118 @@
...
+    OFFENDER_PROCEEDING_ID - not used
+    WORKFLOW_ID - not used
+   */
+
+) {
+  @Column(name = "CREATE_USER_ID", insertable = false, updatable = false)
+  @Generated
+  lateinit var createUsername: String
+
+  @Column(name = "CREATE_DATETIME", insertable = false, updatable = false)
```
mikehalmamoj, line 107: Probably makes no difference if this isn't used anywhere in your tests - but should this also be `@Generated`? Although if it's not used, maybe no need to map the columns? I noticed that modify audit fields aren't included. But I can't keep up with what we think is best practise in this area!

src/main/resources/db/migration/V1_293__ORDER_PURPOSES.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE ORDER_PURPOSES (
```
mikehalmamoj, line 1: I can't see this used anywhere - coming soon?

lbennett-moj: yes - the next one.  I should have mentioned that

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/CourtOrder.kt
lbennett-moj: we tend to have to add these two in for migration, but you are right - I need to check they are all required once migration is stabilised. Not sure what generated is doing - maybe prevents a lateinit warning

mikehalmamoj: Ah I think I know that one! It tells hibernate to get those populate those properties from the DB after a save.

PR #542, opened 2023-12-01
SDIT-1226 sentence purpose, mappings, builders etc

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/SentencePurpose.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.hibernate.annotations.Generated
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.helper.EntityOpen
+import java.io.Serializable
+import java.time.LocalDateTime
+
+@Embeddable
+class SentencePurposeId(
+  var orderId: Long,
+  var orderPartyCode: String, // always 'CRT' on prod
+  var purposeCode: String,
```
lbennett-moj, line 18: this reference data (purposeCode) is awkward to map as part of the embedded key. Any create will validate the reference code.

PR #545, opened 2023-12-06
SDIT-1226 offender sentence, mappings, builders etc

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/CourtEventCharge.kt
```diff
@@ -63,7 +63,7 @@ class CourtEventCharge(
 
   val propertyValue: BigDecimal?,
 
-  val totalPropertyValue: BigDecimal?,
+  val totalPropertyValue: BigDecimal?, // offences taken into consideration button on OCDCCASE
```
andymarke, line 66: is that copy n paste comment from above?

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderSentence.kt
```diff
@@ -29,29 +37,187 @@ data class SentenceId(
...
   )
   @BatchSize(size = 25)
   val calculationType: SentenceCalculationType,
+
   @Column(name = "START_DATE")
   val startDate: LocalDate,
+
+  @ManyToOne(fetch = FetchType.LAZY)
+  @JoinColumn(name = "ORDER_ID")
+  val courtOrder: CourtOrder? = null,
```
andymarke, line 67: interesting it is nullable - a sentence without a court order feels ILLEGAL

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/CourtEventCharge.kt
lbennett-moj: No it's the same pop up

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderSentence.kt
lbennett-moj: yes - plenty of nulls. Will need to understand the meaning of no order Id, but for the mapping this is correct

PR #552, opened 2023-12-12
SDIT-1320 Always write null activity IEP level

mikehalmamoj:
It's no longer used in NOMIS as we don't allocate prisoners to activities there, we do it in DPS


petergphillips (approved):
LGTM

PR #559, opened 2023-12-20
SDIT-1382 create court appearance

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/CourtCase.kt
```diff
@@ -106,7 +106,7 @@ class CourtCase(
   val lidsCombinedCaseId: Int? = null,
 
   @OneToMany(mappedBy = "courtCase", cascade = [CascadeType.ALL], orphanRemoval = true)
-  val courtEvents: MutableList<CourtEvent> = mutableListOf(),
+  var courtEvents: MutableList<CourtEvent> = mutableListOf(),
```
andymarke, line 109: interesting, would have thought it would remain as a val and you just the list contents

andymarke: I see, makes my sense to assign on a POST

andymarke (approved):
lgtm

PR #584, opened 2024-01-17
SDIT-1422 offence result and charge status logic

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingService.kt
```diff
@@ -226,6 +226,7 @@ class SentencingService(
       ).also { courtEvent ->
         request.existingOffenderChargeIds.map { offenderChargeId ->
           getOffenderCharge(offenderChargeId).let { offenderCharge ->
+            val resultCode = offenderCharge.resultCode1?.let { it } ?: courtEvent.outcomeReasonCode
```
petergphillips, line 229: not quite sure what the `let { it }` is doing? Could it be simplified to
```kotlin
offenderCharge.resultCode1 ?: courtEvent.outcomeReasonCode
```

petergphillips (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingService.kt
lbennett-moj: Agreed - must have been a victim of a refactor

PR #586, opened 2024-01-18
Sdit 1396 update event

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingService.kt
```diff
@@ -293,6 +294,51 @@ class SentencingService(
...
   }
 
+  @Audit
+  fun updateCourtAppearance(
+    offenderNo: String,
+    caseId: Long,
+    eventId: Long,
+    request: UpdateCourtAppearanceRequest,
+  ) {
+    findPrisoner(offenderNo).findLatestBooking().let { booking ->
```
andymarke, line 304: I think it is worth thinking of how we can get around using `.findLatestBooking()`  since it means this will never work if they need to update an historic appearance - that you just know is bound to happen on day. You might need some sort of findPrisonerBookingForCourtCase() instead

andymarke (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingService.kt
lbennett-moj: yes - it is on my TODO list - removing this shouldn't be an issue

PR #588, opened 2024-01-19
SDIT-1421 Add offender and staff parties to incident

karenmillermoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/Incident.kt
```diff
@@ -3,58 +3,55 @@ package uk.gov.justice.digital.hmpps.nomisprisonerapi.helper.builders
...
-  /* TODO
   @IncidentPartyDslMarker
-  fun party(
-    role: PartyRole = PartyRole.WITNESS,
+  fun incidentParty(
+    role: String = "VICT",
     offenderBooking: OffenderBooking? = null,
     staff: Staff? = null,
-    comment: String = "They witnessed everything",
+    comment: String? = "They witnessed everything",
```
karenmillermoj, line 25: Not sure why I changed this to String? - will change back in next PR

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/IncidentParty.kt
```diff
@@ -0,0 +1,88 @@
...
+class IncidentParty(
+
+  @EmbeddedId
+  val id: IncidentPartyId,
+
+  @ManyToOne(fetch = FetchType.LAZY, optional = false)
+  @JoinColumn(name = "INCIDENT_CASE_ID", insertable = false, updatable = false)
+  val incident: Incident,
+
+  // Combination of Reference Codes: Staff role code=IR_STF_PART and Offender role code = IR_OFF_PART
```
karenmillermoj, line 39: Because this is a combination of both Staff and Offender role codes, it means that the referenceCode can't be easily mapped.  Currently investigating if a hierarchical "Party" - one for offender and one for staff would mean I could separate them out.

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incidents/IncidentService.kt
```diff
@@ -43,13 +49,28 @@ class IncidentService(
...
-    )
-
-  fun uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.Staff.toStaff() =
-    Staff(
-      staffId = id,
-      firstName = firstName,
-      lastName = lastName,
-      username = accounts.maxByOrNull { it.type }?.username ?: "unknown",
+      staffParties = this.staffParties().map { it.staffParty().toStaff() },
+      offenderParties = this.offenderParties().map { it.offenderParty().toOffender() },
```
karenmillermoj, line 53: These lines look a bit faffy to me - will see if I can improve them/rename in next PR to make it more obvious what's actually happening.

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incidents/IncidentService.kt
```diff
@@ -43,13 +49,28 @@ class IncidentService(
...
     )
 }
+
+private fun Incident.staffParties(): List<IncidentParty> =
+  this.parties.filter { it.isStaffParty() }
+
+private fun uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.Staff.toStaff() =
```
mikehalmamoj, line 60: Can you create an import alias for this instead of full package?

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/IncidentParty.kt
```diff
@@ -0,0 +1,88 @@
...
+    return id == other.id
+  }
+
+  override fun hashCode(): Int = javaClass.hashCode()
+}
+fun IncidentParty.isStaffParty(): Boolean = staff != null
+fun IncidentParty.isOffenderParty(): Boolean = offenderBooking != null
+
+fun IncidentParty.staffParty(): Staff = staff!!
+fun IncidentParty.offenderParty(): Offender = offenderBooking!!.offender
```
mikehalmamoj, line 88: I think the bang-bangs imply that these can't be null? So `isStaffParty` and `isOffenderParty` will always return true?

mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incidents/IncidentService.kt
karenmillermoj: I noticed that too - yep, a better solution

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/IncidentParty.kt
karenmillermoj: yeah - it doesn't - I kinda copied from elsewhere.  They can be null - but "should" only ever be called if they're not null.  Will change as it's a bit confusing.

PR #593, opened 2024-01-23
SDIT-1421 Add IncidentResponse mapping

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/IncidentQuestion.kt
```diff
@@ -15,8 +17,9 @@ import java.time.LocalDateTime
 
 @Embeddable
 class IncidentQuestionId(
-  @Column(name = "INCIDENT_CASE_ID", nullable = false)
```
karenmillermoj, line 18: Changed the mapping to the actual incident to avoid unnecessary access to incident (line 38) and to improve mapping from new IncidentResponse

petergphillips (approved):
LGTM

PR #595, opened 2024-01-24
SDIT-1369 deletion of court event charges/offender charges

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingService.kt
```diff
@@ -299,53 +304,69 @@ class SentencingService(
...
+              val updateCharge = chargesToUpdateMap.getValue(courtEventCharge.id.offenderCharge.id)
+              val resultCode = updateCharge.resultCode1?.let { rs -> lookupOffenceResultCode(rs) }
+              courtEventCharge.offenceDate = updateCharge.offenceDate
+              courtEventCharge.offenceEndDate = updateCharge.offenceEndDate
+              courtEventCharge.mostSeriousFlag = updateCharge.mostSeriousFlag
+              courtEventCharge.offencesCount = updateCharge.offencesCount
+              courtEventCharge.resultCode1 = resultCode
+              courtEventCharge.resultCode1Indicator = resultCode?.dispositionCode
+              with(courtEventCharge.id.offenderCharge) {
+                offenceDate = updateCharge.offenceDate
```
andymarke, line 336: so this is the bit we need to look at? Since I guess the NOMIS rules will be only update from latest appearance date

andymarke (approved):
lgtm

PR #598, opened 2024-01-25
SDIT-1421 Map subclasses for IncidentParty

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/IncidentParty.kt
```diff
@@ -38,14 +42,6 @@ class IncidentParty(
   @Column(name = "PARTICIPATION_ROLE")
```
karenmillermoj, line 42: I can pull this into the subclasses too - missed it!

andymarke (approved):
lgtm

PR #605, opened 2024-02-02
SDIT-1453 Remove prisonIds from adjudication summary

andymarke:
Removed the usage of 'prisonIds' in the AdjudicationService, and its related test cases. This simplifies the adjudication summary response, as we no longer track and include the list of prisons attended during the booking in the response data.


petergphillips (approved):
LGTM

PR #613, opened 2024-02-08
SDIT-1483 Implement workflow logs for offender alerts

andymarke:
Introduced new classes 'AlertWorkFlow' and 'WorkFlowAction', and added necessary SQL migrations, allowing workflow logs to be associated with offender alerts in the 'nomisprisonerapi'. Offender alerts can now hold a workflow with associated logs representing different workflow actions with specific statuses and relevant information. Existing unit tests have been updated to verify the functionality of these changes.


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderAlert.kt
```diff
@@ -104,7 +109,7 @@ class OffenderAlert(
 
   @Column(name = "CREATE_USER_ID", insertable = false, updatable = false)
   @Generated
-  lateinit var createUserId: String
+  lateinit var createUsername: String
```
andymarke, line 113: renamed to be consistent with all other usernames

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/OffenderAlert.kt
```diff
@@ -71,7 +80,16 @@ class OffenderAlertBuilder(
     alertStatus = status,
     commentText = commentText,
     verifiedFlag = verifiedFlag,
-
-  )
```
andymarke, line 75: this apply feels painful when I just want it in the constructor, but can't because it is a one-to-one - not sure if there is a better way

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/WorkFlow.kt
```diff
@@ -0,0 +1,50 @@
...
+import jakarta.persistence.OneToMany
+import jakarta.persistence.SequenceGenerator
+import org.hibernate.Hibernate
+import org.hibernate.annotations.Generated
+import java.time.LocalDateTime
+
+@Entity(name = "WORK_FLOWS")
+@DiscriminatorColumn(name = "OBJECT_CODE")
+@Inheritance
+abstract class WorkFlow(
```
andymarke, line 20: Table is overloaded to point back to different entities based on the OBJECT_CODE

petergphillips (approved):
LGTM

PR #614, opened 2024-02-09
SDIT-1483 Changed WorkFlow to OneToMany

andymarke:
WorkFlow for an Alert changed to OneToMany since there is a small but significant amount of bad data in NOMIS where alert has no workflows or more than one


petergphillips (approved):
LGTM

PR #615, opened 2024-02-12
SDIT-1421 Add prisonId to incident response

karenmillermoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incidents/IncidentResourceIntTest.kt
```diff
@@ -40,72 +39,10 @@ class IncidentResourceIntTest : IntegrationTestBase() {
 
   @BeforeEach
   internal fun createIncidents() {
-    lateinit var question11: QuestionnaireQuestion
```
karenmillermoj, line 43: The config to set up incidents was making it difficult to read (comment from previous PR).  Separated the builders so that questionnaires/staff/offenders used for the incident are created in separate methods  - this makes it much easier to follow

mikehalmamoj (approved):

PR #616, opened 2024-02-12
SDIT-1483 Add alert retrieval with audit data

andymarke:
The changes include a new feature to retrieve alert data by booking ID and alert sequence. An endpoint is added for this feature along with its security and access roles. The code also encompasses creation of new classes to support the alert object and the audit details. This facilitates retrieval of detailed audit information for each alert. Tests are included for validating successful retrieval and various fail cases.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/alerts/AlertsResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/alerts/AlertsResource.kt
@@ -1,0 +1,136 @@
+package uk.gov.justice.digital.hmpps.nomisprisonerapi.alerts
+
+import io.swagger.v3.oas.annotations.Operation
+import io.swagger.v3.oas.annotations.media.Content
+import io.swagger.v3.oas.annotations.media.Schema
+import io.swagger.v3.oas.annotations.responses.ApiResponse
+import org.springframework.http.HttpStatus
+import org.springframework.http.MediaType
+import org.springframework.security.access.prepost.PreAuthorize
+import org.springframework.validation.annotation.Validated
+import org.springframework.web.bind.annotation.GetMapping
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/alerts/AlertsResource.kt
```diff
@@ -0,0 +1,136 @@
...
+  val isVerified: Boolean = false,
+  @Schema(description = "Free format text of person or department that authorised the alert", example = "security")
+  val authorisedBy: String? = null,
+  @Schema(description = "Free format comment")
+  val comment: String? = null,
+  @Schema(description = "Audit data associated with the records")
+  val audit: NomisAudit,
+)
+
+@Schema(description = "The data held in NOMIS the person or system that created this record")
```
andymarke, line 112: keeping audit data separate since I'm not convinced DPS need this even though they have asked for it. Same applies to DSL

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/OffenderAlert.kt
```diff
@@ -45,9 +65,60 @@ class OffenderAlertBuilderRepository(
...
+    modifyDatetime: LocalDateTime?,
+    auditTimestamp: LocalDateTime?,
+    auditUserId: String?,
+    auditModuleName: String?,
+    auditClientUserId: String?,
+    auditClientIpAddress: String?,
+    auditClientWorkstationName: String?,
+    auditAdditionalInfo: String?,
+  ) {
+    jdbcTemplate.update(
```
andymarke, line 87: chosen to do all this by SQL since these fields are set by NOMIS so feels a shame to corrupt the r/w JPA mappings. And also I'm not convinced we will need this code in the future

PR #624, opened 2024-02-16

steverendell:
Update, de/re-activate and change capacity and certification

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/appointments/AppointmentsResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/appointments/AppointmentsResource.kt
@@ -49,7 +49,7 @@ class AppointmentsResource(private val appointmentService: AppointmentService) {
         responseCode = "201",
         description = "Appointment information with created id",
         content = [
-          Content(mediaType = "application/json", schema = Schema(implementation = CreateAppointmentRequest::class)),
+          Content(mediaType = "application/json", schema = Schema(implementation = CreateAppointmentResponse::class)),
         ],
       ),
       ApiResponse(
@@ -82,17 +82,11 @@ class AppointmentsResource(private val appointmentService: AppointmentService) {
     description = "Updates an existing appointment. Requires role NOMIS_APPOINTMENTS",
     requestBody = io.swagger.v3.oas.annotations.parameters.RequestBody(
       content = [
-        Content(mediaType = "application/json", schema = Schema(implementation = CreateAppointmentRequest::class)),
+        Content(mediaType = "application/json", schema = Schema(implementation = UpdateAppointmentRequest::class)),
       ],
     ),
     responses = [
-      ApiResponse(
-        responseCode = "200",
-        description = "Success",
-        content = [
-          Content(mediaType = "application/json", schema = Schema(implementation = CreateAppointmentRequest::class)),
-        ],
-      ),
+      ApiResponse(responseCode = "200", description = "Success"),
       ApiResponse(
         responseCode = "404",
         description = "Event id does not exist",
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/appointments/AppointmentsResource.kt
```diff
@@ -49,7 +49,7 @@ class AppointmentsResource(private val appointmentService: AppointmentService) {
         responseCode = "201",
         description = "Appointment information with created id",
         content = [
-          Content(mediaType = "application/json", schema = Schema(implementation = CreateAppointmentRequest::class)),
+          Content(mediaType = "application/json", schema = Schema(implementation = CreateAppointmentResponse::class)),
```
steverendell, line 52: some swagger typos

petergphillips: odd one, I've found that sometimes you don't need to specify the response and it works it out automatically, but don't really understand when and why that is

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/appointments/AppointmentsResource.kt
```diff
@@ -82,17 +82,11 @@ class AppointmentsResource(private val appointmentService: AppointmentService) {
...
     ),
     responses = [
-      ApiResponse(
-        responseCode = "200",
-        description = "Success",
-        content = [
-          Content(mediaType = "application/json", schema = Schema(implementation = CreateAppointmentRequest::class)),
-        ],
-      ),
+      ApiResponse(responseCode = "200", description = "Success"),
```
petergphillips, line 89: actually you've done my previous comment here anyway 🤷

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/locations/LocationService.kt
```diff
@@ -49,7 +54,152 @@ class LocationService(
...
+      ?: throw NotFoundException("Location with id=$locationId does not exist")
+
+    if (location.deactivateDate != null) {
+      if (location.reactivateDate == null || location.reactivateDate!!.isBefore(location.deactivateDate)) {
+        throw BadDataException("Location with id=$locationId is already deactivated")
+      }
+    }
+
+    location.deactivateDate = LocalDate.now()
+    location.deactivateReason = deactivateRequest.reasonCode?.let {
```
petergphillips, line 125: am I right in thinking they don't have to give a reason for a deactivate request?  Why is that?

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/locations/LocationService.kt
```diff
@@ -49,7 +54,152 @@ class LocationService(
...
+    )
+  }
+
+  @Audit
+  fun reactivateLocation(locationId: Long) {
+    val location = agencyInternalLocationRepository.findByIdOrNull(locationId)
+      ?: throw NotFoundException("Location with id=$locationId does not exist")
+
+    if (location.deactivateDate == null ||
+      (location.reactivateDate != null && !location.reactivateDate!!.isBefore(location.deactivateDate))
```
petergphillips, line 146: strange `!!` here - can never be null as previous condition prevents it?

petergphillips (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/locations/LocationService.kt
steverendell: Looking at existing data there is usually no reason entered

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/locations/LocationService.kt
steverendell: Well good Q ... its a shame but otherwise intellij gives an error "Smart cast to 'LocalDate' is impossible, because 'location.reactivateDate' is a mutable property that could have been changed by this time" i.e. its a var

PR #625, opened 2024-02-19

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/locations/CreateLocationRequest.kt
```diff
@@ -51,7 +51,7 @@ data class CreateLocationRequest(
   )
   val locationCode: String,
 
-  @Schema(description = "Full code hierarchy", example = "MDI-C-3-015")
+  @Schema(description = "Full code hierarchy", required = true, example = "MDI-C-3-015")
```
mikehalmamoj, line 54: You shouldn't ever need required - it defaults from the nullability of the variable

mikehalmamoj (approved):

PR #628, opened 2024-02-20

steverendell:
Correct de/re-activate logic


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/locations/CreateLocationRequest.kt
```diff
@@ -17,7 +17,6 @@ data class CreateLocationRequest(
 
   @Schema(
     description = "Whether a CELL, VISIT room, Kitchen etc (Ref type ILOC_TYPE)",
-    required = true,
```
steverendell, line 20: These are redundant apparently

lbennett-moj (approved):
lgtm

PR #632, opened 2024-02-27
SDIT-1527 Unpin Hibernate after fixing builder.

andymarke:
Builder was referencing an unsaved WorkFlow instance when adding logs - not sure how this previously worked

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderAlert.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderAlert.kt
@@ -142,7 +142,7 @@ class OffenderAlert(
   @Generated
   var auditAdditionalInfo: String? = null
 
-  fun addWorkFlowLog(workActionCode: WorkFlowAction, workFlowStatus: WorkFlowStatus = DONE): WorkFlow {
+  fun addWorkFlowLog(workActionCode: WorkFlowAction, workFlowStatus: WorkFlowStatus = DONE) {
     if (this.workFlows.isEmpty()) this.workFlows.add(AlertWorkFlow(this))
     val workFlow = this.workFlows.first()
     workFlow.logs.add(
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderAlert.kt
```diff
@@ -142,7 +142,7 @@ class OffenderAlert(
   @Generated
   var auditAdditionalInfo: String? = null
 
-  fun addWorkFlowLog(workActionCode: WorkFlowAction, workFlowStatus: WorkFlowStatus = DONE): WorkFlow {
+  fun addWorkFlowLog(workActionCode: WorkFlowAction, workFlowStatus: WorkFlowStatus = DONE) {
```
andymarke, line 145: return no longer used

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/OffenderAlert.kt
```diff
@@ -162,9 +162,10 @@ class OffenderAlertBuilder(
     commentText = commentText,
     verifiedFlag = verifiedFlag,
   ).apply {
-    workFlow = addWorkFlowLog(workActionCode = repository.lookupWorkFLowAction(WorkFlowAction.DATA_ENTRY))
+    addWorkFlowLog(workActionCode = repository.lookupWorkFLowAction(WorkFlowAction.DATA_ENTRY))
   }
     .let { repository.save(it) }
+    .also { workFlow = it.workFlows.first() }
```
andymarke, line 168: set the workFlow instance after it has been saved so the child logs referenced a saved entity - weird this worked before upgrade.

petergphillips (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/OffenderAlert.kt
mikehalmamoj: That is weird. I always thought it was quite nice that Hibernate allowed this - didn't realise it was a bug!

PR #633, opened 2024-02-27

steverendell:
Add attributes and usages


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/locations/CreateLocationRequest.kt
```diff
@@ -24,6 +24,8 @@ data class CreateLocationRequest(
       "POSI", "RESI", "ROOM", "RTU", "SHEL", "SPOR", "SPUR", "STOR", "TABL",
       "TRAI", "TRRM", "VIDE", "VISIT", "WING", "WORK",
     ],
+    // TODO it would be nice to define these values in one place, possibly something like this:
+    // ref = "#/components/schemas/LocationUsageRequest.usageLocationType",
```
steverendell, line 28: I think there is a way to be able to just declare this enum list once and use ref to apply it to fields on other objects

PR #636, opened 2024-03-01
SDIT-1551 In-cell appointments not working

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/appointments/AppointmentService.kt
```diff
@@ -47,8 +47,10 @@ class AppointmentService(
   fun updateAppointment(eventId: Long, dto: UpdateAppointmentRequest) {
     offenderIndividualScheduleRepository.findByIdOrNull(eventId)
       ?.apply {
-        val location = agencyInternalLocationRepository.findByIdOrNull(dto.internalLocationId)
-          ?: throw BadDataException("Room with id=${dto.internalLocationId} does not exist")
+        val location = dto.internalLocationId?.let {
```
mikehalmamoj, line 50: So are they sending null internal location id for in cell? Might be worth testing this in DPS to see what it looks like... especially if they're never null in NOMIS.

mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/appointments/AppointmentService.kt
steverendell: Yep - Dom will be checking!

PR #638, opened 2024-03-01
SDIT-1335 Fix bug preventing attendance being created

mikehalmamoj:
Scenario is:

- allocated with future start date
- deallocated
- allocated again with same future start date 

When the future start date comes around we validate new attendances against the ended allocation and reject the attendance create request - but we should validate against the new allocation and create the attendace


andymarke (approved):
lgtm

PR #645, opened 2024-03-08
SDIT-1582 Handle updates to expired allocations

mikehalmamoj:
Scenario is:
* prisoner allocated to a course in DPS
* later (at least next day), the pay band was changed in DPS
* the same day as the pay band change the prisoner was released
* the next day the deallocation reason was updated in DPS

This caused us to write a duplicate of the expired pay band, so the insert failed with a data integrity issue and the event was rejected.

We now just leave the pay bands alone if the allocation is expired.


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/AllocationService.kt
```diff
@@ -206,7 +206,14 @@ class AllocationService(
...
     val futurePayBand = payBands.firstOrNull { it.id.startDate > today }
+    val allocationExpired = request.endDate?.let { it < today } ?: false
     when {
+      // if the requested pay band is already expired then just make sure the end dates match
+      allocationExpired && activePayBand == null && futurePayBand == null -> {
+        expiredPayBand?.let {
+          it.endDate == request.endDate
+        }
+      }
       // if no active pay band then add a new pay band from today (or the date requested if in the future)
```
petergphillips, line 217: comment doesn't really fit - since this condition is only now triggered if the allocation hasn't expired

petergphillips (approved):
LGTM

PR #651, opened 2024-03-12
SDIT-1590 Fix bug with pay rate sync for future ended rate

mikehalmamoj:
Scenario:
* we have a pay rate that has a future end date
* we receive an update to change the value of the pay rate and remove the end date
* the existing rate's end date isn't updated, so in NOMIS it appears to be active when it's not

The fix is to end the existing pay rate today and start the new one from tomorrow.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/PayRateServiceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/PayRateServiceTest.kt
@@ -409,6 +409,36 @@ class PayRateServiceTest {
     }
 
     @Test
+    fun `should be able to change a future ended pay rate`() {
+      nomisDataBuilder.build {
+        programService {
+          courseActivity = courseActivity {
+            courseSchedule()
+            courseScheduleRule()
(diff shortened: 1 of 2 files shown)


steverendell (approved):
lgtm

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/PayRateServiceTest.kt
```diff
@@ -408,6 +408,36 @@ class PayRateServiceTest {
...
+      }
+      val request = listOf(
+        PayRateRequest("STD", "5", BigDecimal(0.1)),
+      )
+      val newPayRates = payRatesService.buildNewPayRates(request, courseActivity)
+
+      assertThat(newPayRates.size).isEqualTo(2)
+      // old rate should now expire today
+      with(newPayRates.findRate("STD", "5", expired = true)) {
+        assertThat(halfDayRate).isCloseTo(BigDecimal(1.41), within(BigDecimal(0.001)))
```
petergphillips, line 430: this looks a bit odd - why the `closeTo`?  If you called `BigDecimal` constructor with a string instead then you should be able to use [isEqualByComparingTo](https://joel-costigliola.github.io/assertj/core-8/api/org/assertj/core/api/AbstractBigDecimalAssert.html#isEqualByComparingTo-java.lang.String-)

PR #654, opened 2024-03-13
SDIT-1586 Audit call failure results in DB inconsistency

steverendell:
Fix based on initialising the connection


andymarke (approved):
LGTM

PR #666, opened 2024-03-22
SDIT-1612 Implement alert filtering by date range

andymarke:
Introduced the ability to fetch alerts by filtering through a specific date range. The backend API now supports retrieving alerts from a 'fromDate' to a 'toDate'. Enhancements were made to the AlertsService and OffenderAlertRepository to facilitate this feature. Corresponding tests were also added. Changes also include an endpoint creation for retrieving alert Ids by applying a filter.


steverendell (approved):
lgtm

PR #669, opened 2024-04-02
SDIT-1625 Endpoint to retrieve merge details for prisoner

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/MergeTransaction.kt
```diff
@@ -0,0 +1,84 @@
...
+  val requestDate: LocalDateTime,
+
+  @Column(name = "REQUEST_STATUS_CODE", nullable = false)
+  val requestStatusCode: String,
+
+  @Column(name = "TRANSACTION_SOURCE", nullable = false)
+  val transactionSource: String,
+
+  @Column(name = "OFFENDER_BOOK_ID_1")
+  val offenderBookId1: Long,
```
andymarke, line 32: most columns on DB are nullable but no rows in live have an nulls

lbennett-moj (approved):

PR #671, opened 2024-04-03
SDIT-1623 Add initial CSIP files for migration

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/CSIPReport.kt
```diff
@@ -0,0 +1,110 @@
...
+import org.hibernate.Hibernate
+import org.hibernate.annotations.Generated
+import org.hibernate.annotations.JoinColumnOrFormula
+import org.hibernate.annotations.JoinColumnsOrFormulas
+import org.hibernate.annotations.JoinFormula
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.helper.EntityOpen
+import java.time.LocalDate
+import java.time.LocalDateTime
+
+@Entity
```
karenmillermoj, line 23: I've currently added minimal data for a CSIP report and associated plan.  Additional fields and other classes still to be added (separately, to avoid BIG PR)

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/csip/CSIPResource.kt
```diff
@@ -0,0 +1,211 @@
...
+  val location: CodeDescription,
+  @Schema(description = "The Area of work where the incident took place aka function")
+  val areaOfWork: CodeDescription,
+
+  @Schema(description = "CSIP Plans")
+  val plans: List<Plan>,
+)
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+data class Staff(
```
andymarke, line 181: I couldn't see where Staff and Offender is used

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/csip/CSIPService.kt
```diff
@@ -0,0 +1,90 @@
...
+    offenderNo = offender.nomsId,
+    bookingId = offenderBooking?.bookingId,
+    type = type.toCodeDescription(),
+    location = location.toCodeDescription(),
+    areaOfWork = areaOfWork.toCodeDescription(),
+    plans = plans.map { it.toPlanResponse() },
+
+  )
+
+private fun JPAStaff.toStaff() =
```
andymarke, line 70: I guess these are just cut & paste things?

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/CSIPReport.kt
```diff
@@ -0,0 +1,110 @@
...
+data class CSIPReport(
+  @Id
+  @Column(name = "CSIP_ID")
+  @SequenceGenerator(name = "CSIP_ID", sequenceName = "CSIP_ID", allocationSize = 1)
+  @GeneratedValue(generator = "CSIP_ID")
+  val id: Long = 0,
+
+  @ManyToOne(fetch = FetchType.LAZY)
+  @JoinColumn(name = "OFFENDER_BOOK_ID")
+  val offenderBooking: OffenderBooking?,
```
andymarke, line 35: Is this really optional? Feels weird to have CSIP with no associated booking

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/csip/CSIPResourceIntTest.kt
```diff
@@ -0,0 +1,211 @@
...
+  private lateinit var reportingStaff: Staff
+  private lateinit var offender: Offender
+  private lateinit var csip1: CSIPReport
+  private lateinit var csip2: CSIPReport
+  private lateinit var csip3: CSIPReport
+
+  @BeforeEach
+  internal fun createCSIPReports() {
+    nomisDataBuilder.build {
+      setUpStaffAndOffender()
```
andymarke, line 35: not sure extracting this to a method makes it more readable - but this is a personal style thing

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/NomisData.kt
```diff
@@ -372,6 +400,16 @@ interface NomisDataDsl {
     dsl: IncidentDsl.() -> Unit = {},
   ): Incident
 
+  @CSIPReportDslMarker
+  fun csipReport(
+    offender: Offender,
```
andymarke, line 405: I wonder if you could do this 
`offender: Offender = offenderBooking.offender`

it feels like you should be able to default at least of them. If you supply a booking you shouldn't really need to supply and offender

Maybe I don't undertstand why booking is optional in the JPA

andymarke: Equally, could a csipReport be a child of an offender (or for that matter a booking) rather then a top level entity?
In  fact wouldn't that make more sense or can you create a CSIP with no relationship with prisoner??

andymarke (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/csip/CSIPResource.kt
karenmillermoj: yet to come

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/csip/CSIPService.kt
karenmillermoj: nope - they're coming shortly

src/main/resources/db/migration/data/R__1_24__REFERENCE_DOMAINS.sql
```diff
@@ -173,6 +173,26 @@ VALUES ('COUNTY', 'County', 'ACTIVE', 'ADMIN', 'OMS', NULL, NULL, NULL);
 INSERT INTO REFERENCE_DOMAINS (DOMAIN, DESCRIPTION, DOMAIN_STATUS, OWNER_CODE, APPLN_CODE, OLD_CODE_TABLE, PARENT_DOMAIN, CODE_LENGTH)
 VALUES ('CRT_RESULT', 'The result of the court event', 'ACTIVE', 'USER', 'LEGAL', NULL, NULL, NULL);
 INSERT INTO REFERENCE_DOMAINS (DOMAIN, DESCRIPTION, DOMAIN_STATUS, OWNER_CODE, APPLN_CODE, OLD_CODE_TABLE, PARENT_DOMAIN, CODE_LENGTH)
```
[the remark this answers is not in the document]
karenmillermoj: ok good idea, will take a look

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/NomisData.kt
karenmillermoj: I think you're right about child of offender.  Looks like the nullability of the offenderBooking means it can't be part of that (although no idea why both are on the table??)
Will make child of offender (obvious, now you've mentioned it)

karenmillermoj: Will address issues in separate PR

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/CSIPReport.kt
karenmillermoj: Well, it's actually nullable, but there are no offenders with this as null.  I'll check the db - maybe it's an oversight on the nullability of fields

PR #677, opened 2024-04-08
SDIT-1654 get court appearance endpoint

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingService.kt
```diff
@@ -275,6 +275,11 @@ class SentencingService(
     }
   }
 
+  fun getCourtAppearance(id: Long, offenderNo: String): CourtEventResponse {
+    findPrisoner(offenderNo).findLatestBooking()
```
andymarke, line 279: Not sure I would bother with this call just to get the correct 404

andymarke (approved):
Lgtm

PR #683, opened 2024-04-10
SDIT-1645 Amended /prisoners/ids to return all prisoners not just active ones.

andymarke:
…

The endpoint has an unfortunate parameter of `active` where setting this false means return active and inactive prisoners.

A prisoner is defined as someone at some-point that had a booking.

Also amened DSL to allow adding of Alias offender records.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderBooking.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderBooking.kt
@@ -160,3 +160,4 @@ data class OffenderBooking(
 }
 
 fun OffenderBooking.hasBeenReleased() = !this.active && this.inOutStatus == "OUT"
+fun OffenderBooking.status() = "${if (this.active) "ACTIVE" else "INACTIVE"} $inOutStatus"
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderBooking.kt
```diff
@@ -160,3 +160,4 @@ data class OffenderBooking(
 }
 
 fun OffenderBooking.hasBeenReleased() = !this.active && this.inOutStatus == "OUT"
+fun OffenderBooking.status() = "${if (this.active) "ACTIVE" else "INACTIVE"} $inOutStatus"
```
andymarke, line 163: will only be used for analytics so use prison-api convention due familiarity

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/Offender.kt
```diff
@@ -71,6 +83,22 @@ class OffenderBuilder(
     .also { it.rootOffenderId = it.id }
     .also { offender = it }
 
+  private fun buildAlias(
```
mikehalmamoj, line 86: I was expecting this to be in a different file for the AliasDsl... is that because Alias is the same as Offender?

mikehalmamoj (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/Offender.kt
andymarke: Yes - a sort of special case. And Offender has an alias which is really just another Offender - so really you are creating an offender from within an offender.

But that is confusing so called it "alias" which is closer to what we know it as

PR #691, opened 2024-04-22
SDIT-1659 Previous alert of same type rules

andymarke:
DPS Rules defined as:
Identical datetime logic (V1):
If more than one of the most recent alerts have the same datetime. 

We use the following rules: 

* If one active alert exists in this subset, we surface that active alert.
*  If more than one identical active alerts exist, we surface the active alert with the most recent audit datetime. 
* If no active alerts exist in the subset, we surface the alert with the most recent audit datetime. 
* If audit datetime is the same for more two or more alerts in either point 1 or 2 then we surface all alerts with the same audit datetime.


lbennett-moj (approved):
very thorough testing!

PR #692, opened 2024-04-22
SDIT-1659 Previous alert of same type rules

andymarke:
DPS Rules change
* In addition, booking sequence (i.e. most recent first) trumps alert date

Refactored unit test alert creation so that it points back at booking

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/alerts/AlertsResourceIntTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/alerts/AlertsResourceIntTest.kt
@@ -486,16 +486,6 @@ class AlertsResourceIntTest : IntegrationTestBase() {
               commentText = null,
             )
             alert(
-              sequence = 4,
-              alertCode = "P1",
-              typeCode = "P",
-              date = LocalDate.parse("2017-07-19"),
-              authorizePersonText = "Security Team",
-              verifiedFlag = true,
-              status = ACTIVE,
-              commentText = "MAPPA on middle booking but old date",
-            )
-            alert(
               sequence = 5,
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/alerts/AlertsResourceIntTest.kt
```diff
@@ -485,16 +485,6 @@ class AlertsResourceIntTest : IntegrationTestBase() {
               status = ACTIVE,
               commentText = null,
             )
-            alert(
```
andymarke, line 488: No logic would selected this alert, so removed this alert to make test clearer

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/alerts/AlertsServiceTest.kt
```diff
@@ -50,9 +50,7 @@ class AlertsServiceTest {
       fun setUp() {
         whenever(offenderBookingRepository.findAllByOffenderNomsId("A1234KT")).thenReturn(
           listOf(
-            booking(
-              listOf(),
-            ),
+            booking(),
           ),
```
andymarke, line 54: no longer pass in a list of alerts but `apply` so we can grab the booking and set on the alert.
That has a big impact on how the tests data is created unfortunately

andymarke: So the changes below are essentially more the duplicate alerts to a single booking and then a new test to show more recent bookings take priority over older bookings . But it is hard to see any of that in the diff

mikehalmamoj (approved):

PR #700, opened 2024-04-29
SDIT-1623 Rename CSIP Investigation class to avoid name clash

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/csip/CSIPResource.kt
```diff
@@ -201,7 +201,7 @@ data class CSIPResponse(
   val saferCustodyScreening: SaferCustodyScreening,
 
   @Schema(description = "Investigation details of the incident")
-  val investigation: Investigation,
+  val investigation: InvestigationDetails,
```
karenmillermoj, line 204: Investigation class name already used as a separate classso when documentation created, it can't cope with name clash.  There is a setting to ensure uniqueness but I didn't get very far with this so just renamed.

lbennett-moj (approved):

PR #703, opened 2024-05-01
SDIT-1704 Add prison id to sentence adjustments endpoint

andymarke:
This will be passed to the DPS service so they don't need to make an additional call to retrieve this data.


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingAdjustmentService.kt
```diff
@@ -271,6 +271,7 @@ private fun OffenderKeyDateAdjustment.toAdjustmentResponse() =
     active = this.active,
     offenderNo = this.offenderBooking.offender.nomsId,
     hasBeenReleased = this.offenderBooking.hasBeenReleased(),
+    prisonId = this.offenderBooking.location?.id ?: "OUT",
   )
```
andymarke, line 275: oddly there are 5 null records in production hence this nonsense

steverendell (dismissed):
lgtm

steverendell (approved):

PR #716, opened 2024-05-14
SDIT-1730 Get Alert indicates if this is a relevant alert from previous booking

andymarke:
…

Synchronisation service needs to know if this alert should be pushed to DPS if it does not exist. This is determined by the rules that were used to determine if it had been migrated. Typically, an alert that is missing from latest booking.


petergphillips (approved):
LGTM

PR #719, opened 2024-05-17
SDIT-1421 Add additional required fields to incidents

karenmillermoj:


mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incidents/IncidentResource.kt
```diff
@@ -242,6 +278,10 @@ data class Question(
   val question: String,
   @Schema(description = "List of Responses to this question")
   val answers: List<Response> = listOf(),
+  @Schema(description = "The date and time the question was created")
+  val createDateTime: LocalDateTime,
+  @Schema(description = "The username of the person who created the question")
+  val createdBy: String,
```
xoen, line 284: Hi @karenmillermoj I notice `Question` and `History` don't have the `lastModifiedDateTime`/`lastModifiedBy` fields unlike the other classes.

I don't think it's a problem on our end but I'm just double checking in case I'm missing something :)?

PR #731, opened 2024-05-23
SDIT-1778 Modify user set via a Oracle proxy connection

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/alerts/AlertsResource.kt
```diff
@@ -465,7 +466,7 @@ class AlertsResource(
     alertSequence: Long,
     @RequestBody @Valid
     request: UpdateAlertRequest,
-  ): AlertResponse = alertsService.updateAlert(bookingId, alertSequence, request)
+  ): AlertResponse = proxy(request.updateUsername) { alertsService.updateAlert(bookingId, alertSequence, request) }
```
andymarke, line 470: need to set the MDC state prior to the transaction starting

PR #735, opened 2024-05-28
SDIT-1378 Add reconciliation methods to get incident locations and incident counts

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incidents/IncidentResource.kt
```diff
@@ -128,6 +132,67 @@ class IncidentResource(private val incidentService: IncidentService) {
   fun getIncident(
     @Schema(description = "Incident id") @PathVariable incidentId: Long,
   ) = incidentService.getIncident(incidentId)
+
```
karenmillermoj, line 135: I will also be getting a list of open incident ids per prison (to be able to then compare open incident details) but this will be a separate PR

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/IncidentRepository.kt
```diff
@@ -37,4 +39,26 @@ interface IncidentRepository : CrudRepository<Incident, Long>, JpaSpecificationE
...
+
+  @Query(
+    """
+      select new uk.gov.justice.digital.hmpps.nomisprisonerapi.incidents.IncidentsCount(
+        coalesce(sum(case when incident.status.code in (:openStatusValues) THEN 1 else 0 end), 0),
+        coalesce(sum(case when incident.status.code in (:closedStatusValues) THEN 1 else 0 end), 0)
+      )
+      from Incident incident 
+      where incident.location.id = :agencyLocationId
+      order by incident.location.id
```
karenmillermoj, line 60: I don't need the order - will remove in next PR

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incidents/IncidentResource.kt
```diff
@@ -141,8 +206,8 @@ data class IncidentResponse(
   val title: String?,
   @Schema(description = "The incident details")
   val description: String?,
-  @Schema(description = "Prison where the incident occurred")
-  val prison: CodeDescription,
+  @Schema(description = "Agency Location where the incident occurred")
+  val location: CodeDescription,
```
steverendell, line 210: I hate to say it but maybe 'location' seems a bit too easy to confuse with locations (as in within a prison), maybe something like 'agency' and agencyId instead of prisonId is better?

steverendell (approved):
lgtm with comment

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incidents/IncidentResource.kt
karenmillermoj: funny- I toyed with this - I will change to agencyId in separate PR - I found an agencyLocation that wasn't a prison (it was PECS) so  leaving as prison didn't make sense anymore.

PR #736, opened 2024-05-28
SDIT-1753 Remove status from PrisonerIds and add active prisoner endpoint

andymarke:
The status property was removed from the PrisonerIds class as it was no longer used. A new endpoint, "/prisoners/ids/active", was added to allow fetching identifiers of all active prisoners this replaces the deprecated version which will be removed soon


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisoners/PrisonerService.kt
```diff
@@ -25,7 +24,7 @@ class PrisonerService(
 ) {
   fun findAllActivePrisoners(pageRequest: Pageable): Page<PrisonerIds> {
     return bookingRepository.findAll(ActiveBookingsSpecification(), pageRequest)
-      .map { PrisonerIds(bookingId = it.bookingId, offenderNo = it.offender.nomsId, status = it.status()) }
```
andymarke, line 28: the only usage was removed last week (which was just for analytics)

mikehalmamoj (approved):

PR #739, opened 2024-05-28
SDIT-1788: ⬆️ Upgrade to latest spring boot etc.

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderSentence.kt
```diff
@@ -57,7 +56,6 @@ data class OffenderSentence(
       JoinColumn(name = "SENTENCE_CATEGORY", referencedColumnName = "SENTENCE_CATEGORY"),
     ],
   )
-  @BatchSize(size = 25)
```
petergphillips, line 60: needed as this is now an error in hibernate 6.5

andymarke (approved):
LGTM

PR #740, opened 2024-05-29
SDIT-1785 Add ability to create alert codes and types

andymarke:
Updated AlertType and AlertCode classes to allow adding additional attributes. Also, added methods 'createAlertCode' and 'createAlertType' in AlertsResource class to create new alert codes/types. Added new service AlertsReferenceDataService to handle the creation of alert codes and types.


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/ReferenceCode.kt
```diff
@@ -36,6 +36,7 @@ abstract class ReferenceCode(
   open var sequence: Int? = 0,
 
   open var parentCode: String? = null,
+  open var parentDomain: String? = null,
```
andymarke, line 40: need for ALERT_CODE since they have a parent of ALERT ( the type)

PR #750, opened 2024-06-06
SDIT-1801 Create Casenotes JPA mappings in Nomis api

steverendell:


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/OffenderBooking.kt
```diff
@@ -167,7 +169,7 @@ interface BookingDsl {
     beginDate: LocalDate = LocalDate.now(),
     caseSequence: Int = 1,
     caseInfoNumber: String? = "AB1",
-    courtId: String = "COURT1",
+    prisonId: String = "COURT1",
```
mikehalmamoj, line 172: Should you also change the value from `COURT1`?

mikehalmamoj (approved):

PR #760, opened 2024-06-17
SDIT-1660 OffenderBooking/Offender JPA model cascading delete error

steverendell:
Provide a correct prisoner 'get all bookings' function and deprecate the 'bookings' list

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
@@ -91,10 +91,6 @@ data class Offender(
 
   @Column(name = "LAST_NAME_ALPHA_KEY")
   val lastNameAlphaKey: String? = null,
-
-  //    @OneToMany(mappedBy = "offender", cascade = CascadeType.ALL, fetch = FetchType.LAZY)
-  //    @Where(clause = "OWNER_CLASS = '"+OffenderAddress.ADDR_TYPE+"'")
-  //    private List<OffenderAddress> addresses = new ArrayList<>();
 ) {
   val middleNames: String
     get() = StringUtils.trimToNull(StringUtils.trimToEmpty(middleName) + " " + StringUtils.trimToEmpty(middleName2))
@@ -102,7 +98,7 @@ data class Offender(
   fun getAllBookings(): MutableList<OffenderBooking>? = rootOffender?.allBookings
 
   fun latestBooking(): OffenderBooking =
-    getAllBookings()?.firstOrNull { it.active } ?: throw IllegalStateException("Offender has no active bookings")
+    getAllBookings()?.firstOrNull { it.bookingSequence == 1 } ?: throw IllegalStateException("Offender has no active bookings")
 
   override fun equals(other: Any?): Boolean {
     if (this === other) return true
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
@@ -80,6 +80,9 @@ data class Offender(
   @OneToMany(mappedBy = "offender", cascade = [CascadeType.ALL], fetch = FetchType.LAZY, orphanRemoval = true)
   val bookings: MutableList<OffenderBooking> = mutableListOf(),
 
+  @OneToMany(mappedBy = "rootOffender", fetch = FetchType.LAZY)
+  private val allBookings: MutableList<OffenderBooking> = mutableListOf(),
+
   @OneToMany(mappedBy = "offender", cascade = [CascadeType.ALL], fetch = FetchType.LAZY)
   val identifiers: List<OffenderIdentifier> = ArrayList(),
 
@@ -120,6 +123,10 @@ data class Offender(
   val middleNames: String
     get() = StringUtils.trimToNull(StringUtils.trimToEmpty(middleName) + " " + StringUtils.trimToEmpty(middleName2))
 
+  fun getAllBookings(): MutableList<OffenderBooking>? {
+    return rootOffender?.allBookings
+  }
+
   override fun equals(other: Any?): Boolean {
     if (this === other) return true
     if (other == null || Hibernate.getClass(this) != Hibernate.getClass(other)) return false
@@ -128,4 +135,7 @@ data class Offender(
   }
 
   override fun hashCode(): Int = javaClass.hashCode()
+
+  override fun toString(): String =
+    "${javaClass.simpleName}(id = $id, nomsId=$nomsId, firstName = $firstName, lastName = $lastName)"
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
@@ -123,9 +123,10 @@ data class Offender(
   val middleNames: String
     get() = StringUtils.trimToNull(StringUtils.trimToEmpty(middleName) + " " + StringUtils.trimToEmpty(middleName2))
 
-  fun getAllBookings(): MutableList<OffenderBooking>? {
-    return rootOffender?.allBookings
-  }
+  fun getAllBookings(): MutableList<OffenderBooking>? = rootOffender?.allBookings
+
+  fun latestBooking(): OffenderBooking =
+    getAllBookings()?.firstOrNull { it.active } ?: throw IllegalStateException("Offender has no active bookings")
 
   override fun equals(other: Any?): Boolean {
     if (this === other) return true
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
@@ -76,36 +76,12 @@ data class Offender(
   @JoinColumn(name = "ROOT_OFFENDER_ID", updatable = false, insertable = false)
   var rootOffender: Offender? = null,
 
-  @Deprecated("Do not use; only some of the bookings may be associated with the root or any particular alias")
-  @OneToMany(mappedBy = "offender", cascade = [CascadeType.ALL], fetch = FetchType.LAZY, orphanRemoval = true)
-  val bookings: MutableList<OffenderBooking> = mutableListOf(),
-
-  @OneToMany(mappedBy = "rootOffender", fetch = FetchType.LAZY)
+  // CAUTION: this list is only populated for the root offender; normally the function getAllBookings() below should be used instead
+  @OneToMany(mappedBy = "rootOffender", cascade = [CascadeType.ALL], fetch = FetchType.LAZY, orphanRemoval = true)
   private val allBookings: MutableList<OffenderBooking> = mutableListOf(),
 
   @OneToMany(mappedBy = "offender", cascade = [CascadeType.ALL], fetch = FetchType.LAZY)
   val identifiers: List<OffenderIdentifier> = ArrayList(),
-
-  //    @ManyToOne(fetch = FetchType.LAZY)
-  //    @JoinColumnsOrFormulas(value = {
-  //        @JoinColumnOrFormula(formula = @JoinFormula(value = "'" + ETHNICITY + "'", referencedColumnName = "domain")),
-  //        @JoinColumnOrFormula(column = @JoinColumn(name = "RACE_CODE", referencedColumnName = "code"))
-  //    })
-  //    private Ethnicity ethnicity;
-  //
-  //    @ManyToOne(fetch = FetchType.LAZY)
-  //    @JoinColumnsOrFormulas(value = {
-  //        @JoinColumnOrFormula(formula = @JoinFormula(value = "'" + TITLE + "'", referencedColumnName = "domain")),
-  //        @JoinColumnOrFormula(column = @JoinColumn(name = "TITLE", referencedColumnName = "code"))
-  //    })
-  //    private Title title;
-  //
-  //    @ManyToOne(fetch = FetchType.LAZY)
-  //    @JoinColumnsOrFormulas(value = {
-  //        @JoinColumnOrFormula(formula = @JoinFormula(value = "'" + SUFFIX + "'", referencedColumnName = "domain")),
-  //        @JoinColumnOrFormula(column = @JoinColumn(name = "SUFFIX", referencedColumnName = "code"))
-  //    })
-  //    private Suffix suffix;
 
   @Column(name = "LAST_NAME_KEY", nullable = false)
   var lastNameKey: String? = null,
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderBookingRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderBookingRepository.kt
@@ -27,14 +27,6 @@ interface OffenderBookingRepository :
   )
   fun findLatestByOffenderNomsId(@Param("nomsId") nomsId: String): OffenderBooking?
 
-  @Query(
-    """
-    select booking from OffenderBooking booking 
-        join Offender offender on booking.rootOffender = offender 
-        where offender.nomsId = :nomsId 
-        and offender.rootOffender = offender
-        """,
-  )
   fun findAllByOffenderNomsId(@Param("nomsId") nomsId: String): List<OffenderBooking>
 
   fun findOneByOffenderNomsIdAndBookingSequence(nomsId: String, sequence: Int): OffenderBooking?
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderRepository.kt
@@ -23,7 +23,9 @@ interface OffenderRepository : JpaRepository<Offender, Long>, JpaSpecificationEx
 
   @Query(
     value = """
-        select o.nomsId as nomsId, ob.bookingId as bookingId, ob.active as active, ob.inOutStatus as inOutStatus from OffenderBooking ob join ob.offender.rootOffender o where ob.rootOffender = o.rootOffender and ob.bookingSequence = 1
+      select o.nomsId as nomsId, ob.bookingId as bookingId, ob.active as active, ob.inOutStatus as inOutStatus 
+        from OffenderBooking ob join ob.offender o 
+        where ob.bookingSequence = 1
     """,
     countQuery = """
       select count(ob) from OffenderBooking ob where ob.bookingSequence = 1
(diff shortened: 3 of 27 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
```diff
@@ -128,4 +135,7 @@ data class Offender(
   }
 
   override fun hashCode(): Int = javaClass.hashCode()
+
+  override fun toString(): String =
+    "${javaClass.simpleName}(id = $id, nomsId=$nomsId, firstName = $firstName, lastName = $lastName)"
```
steverendell, line 140: Prevent recursion

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderBookingRepository.kt
```diff
@@ -27,14 +27,6 @@ interface OffenderBookingRepository :
...
 
-  @Query(
-    """
-    select booking from OffenderBooking booking 
-        join Offender offender on booking.rootOffender = offender 
-        where offender.nomsId = :nomsId 
-        and offender.rootOffender = offender
-        """,
-  )
   fun findAllByOffenderNomsId(@Param("nomsId") nomsId: String): List<OffenderBooking>
```
steverendell, line 30: no longer needs explicit query as the rootOffender part is incorrect

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderRepository.kt
```diff
@@ -23,7 +23,9 @@ interface OffenderRepository : JpaRepository<Offender, Long>, JpaSpecificationEx
 
   @Query(
     value = """
-        select o.nomsId as nomsId, ob.bookingId as bookingId, ob.active as active, ob.inOutStatus as inOutStatus from OffenderBooking ob join ob.offender.rootOffender o where ob.rootOffender = o.rootOffender and ob.bookingSequence = 1
+      select o.nomsId as nomsId, ob.bookingId as bookingId, ob.active as active, ob.inOutStatus as inOutStatus 
+        from OffenderBooking ob join ob.offender o 
+        where ob.bookingSequence = 1
```
steverendell, line 28: Just formatting

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/Offender.kt
```diff
@@ -112,17 +120,22 @@ class OffenderBuilder(
...
         agencyLocationCode = agencyLocationId,
         bookingBeginDate = bookingBeginDate,
         active = active,
         inOutStatus = inOutStatus,
         youthAdultCode = youthAdultCode,
         livingUnitId = livingUnitId,
       )
-        .also { offender.bookings += it }
+        .also {
+          println("Created $it for ${it.offender}")
```
steverendell, line 135: whoops  -  will remove the print

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/Offender.kt
```diff
@@ -112,17 +120,22 @@ class OffenderBuilder(
...
+          aliasOffender.bookings += it
+          rootOffender.getAllBookings()?.add(it)
+        }
```
steverendell, line 138: Although the lists are correct when retrieved from the db, this is needed so that they are initialised in the records returned from the dsl functions

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/OffenderBookingBuilder.kt
```diff
@@ -46,10 +49,6 @@ class OffenderBookingBuilder(
     this.visits = arrayOf(*visitBuilder).asList()
     return this
   }
-  fun withIncentives(vararg incentiveBuilder: LegacyIncentiveBuilder): OffenderBookingBuilder {
-    this.incentives = arrayOf(*incentiveBuilder).asList()
-    return this
-  }
```
steverendell, line 52: not used

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/Repository.kt
```diff
@@ -136,44 +136,41 @@ class Repository(
...
-                it.build(it.person, leadVisitor = it.leadVisitor, visit)
-              } + visitBuilder.visitOutcome.build(visit),
-            )
-            visit
-          },
-        )
-
-        booking
-      },
-    )
```
steverendell, line 176: just removed the 'addAll' as the list is now added to inside the bookingBuilder.build()

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisonerprofile/PrisonerProfileIntTest.kt
```diff
@@ -143,13 +143,13 @@ class PrisonerProfileIntTest : IntegrationTestBase() {
...
+            booking = booking(bookingBeginDate = today.atStartOfDay()) {
+              physicalAttributes(170, null, null, 70, null)
+            }
+            oldBooking = booking(bookingBeginDate = today.minusDays(2).atStartOfDay()) {
               physicalAttributes(180, null, null, 80, null)
               release(date = yesterday.atStartOfDay())
             }
-            booking = booking(bookingSequence = 1, bookingBeginDate = today.atStartOfDay()) {
-              physicalAttributes(170, null, null, 70, null)
-            }
```
steverendell, line 152: Swapped these so there was no need for the bookingSequence setting

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisoners/PrisonersResourceIntTest.kt
```diff
@@ -22,6 +24,122 @@ class PrisonersResourceIntTest : IntegrationTestBase() {
   @Autowired
   private lateinit var nomisDataBuilder: NomisDataBuilder
 
+  @AfterEach
+  fun cleanUp() {
+    repository.deleteOffenders()
+  }
```
steverendell, line 30: introduced a global prisoner cleanUp

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisoners/PrisonersResourceIntTest.kt
```diff
@@ -22,6 +24,122 @@ class PrisonersResourceIntTest : IntegrationTestBase() {
...
+        activeBooking = booking {}
+        oldBooking = booking(active = false)
+
+        aliasOffender1 = alias(lastName = "ALIAS1") {
+          aliasBooking = booking(active = false)
+          oldAliasBooking = booking(active = false)
+        }
+        aliasOffender2 = alias(lastName = "ALIAS2")
+      }
+    }
```
mikehalmamoj, line 53: The data produced by nomisDataBuilder isn’t correct… I guess that’s why you then load them from the database? That feels like a hack to work around the fact they’re not right. For example aliasOffender1.allBookings is empty before loading from the repository.

I think after the call to nomisDataBuilder any data produced should be in a correct state - maybe that means that the repository load is done within the builders? But expecting people to reload from the database isn’t right.

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisoners/PrisonersResourceIntTest.kt
```diff
@@ -22,6 +24,122 @@ class PrisonersResourceIntTest : IntegrationTestBase() {
...
+    lateinit var oldAliasBooking: OffenderBooking
+
+    nomisDataBuilder.build {
+      rootOffender = offender(nomsId = "A1234TT") {
+        activeBooking = booking {}
+        oldBooking = booking(active = false)
+
+        aliasOffender1 = alias(lastName = "ALIAS1") {
+          aliasBooking = booking(active = false)
+          oldAliasBooking = booking(active = false)
```
mikehalmamoj, line 49: We still have a bug in the builders such that we can add aliases to aliases... e.g. we can do
```
aliasOffender1 = alias(lastName = "ALIAS1") {
  aliasOffenderNested1 = alias(lastName = "ALIASNESTED") {
    aliasOffenderNested2 = alias(lastName = "ALIASNESTEDNESTED") {
...
```
This is why we were advocating for adding an `AliasDSL` - this should be a copy of `OffenderDSL` but without the `alias` method (thus preventing nested aliases).

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
```diff
@@ -80,6 +80,9 @@ data class Offender(
   @OneToMany(mappedBy = "offender", cascade = [CascadeType.ALL], fetch = FetchType.LAZY, orphanRemoval = true)
   val bookings: MutableList<OffenderBooking> = mutableListOf(),
 
+  @OneToMany(mappedBy = "rootOffender", fetch = FetchType.LAZY)
+  private val allBookings: MutableList<OffenderBooking> = mutableListOf(),
+
```
mikehalmamoj, line 85: We still have the `@Deprecated` bookings above - I don't think we need to leave deprecated code lying around for ourselves to unpick later when we've forgotten what's going on. Can we get rid of this now?

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/Offender.kt
```diff
@@ -112,17 +120,22 @@ class OffenderBuilder(
   ) = bookingBuilderFactory.builder()
     .let { builder ->
       builder.build(
-        offender = offender,
-        // if you want multiple bookings then create them with latest booking first so it gets seq 1 like in Nomis
-        bookingSequence = bookingSequence ?: (offender.bookings.size + 1),
+        offender = aliasOffender,
+        // if you want multiple bookings then create them with latest booking first so it gets seq 1 like in Nomis,
+        // or override the sequences
+        bookingSequence = bookingSequence ?: nextBookingSequence++,
```
mikehalmamoj, line 126: I thought you'd removed the `bookingSequence` parameter from the test I added it to - so can we remove from the DSL entirely if it's not needed?

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/Offender.kt
```diff
@@ -62,7 +62,9 @@ class OffenderBuilder(
   private val repository: OffenderBuilderRepository,
   private val bookingBuilderFactory: BookingBuilderFactory,
 ) : OffenderDsl {
-  private lateinit var offender: Offender
+  private lateinit var rootOffender: Offender
+  private lateinit var aliasOffender: Offender
```
mikehalmamoj, line 66: I was kind of expecting this to be a list of aliases rather than a single one given you can call the `alias` method multiple times.

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/integration/IntegrationTestBase.kt
```diff
@@ -33,4 +33,4 @@ abstract class IntegrationTestBase {
 }
 
 fun Offender.latestBooking(): OffenderBooking =
-  this.bookings.firstOrNull { it.active } ?: throw IllegalStateException("Offender has no active bookings")
+  this.getAllBookings()?.firstOrNull { it.active } ?: throw IllegalStateException("Offender has no active bookings")
```
mikehalmamoj, line 36: I wonder if we should move `Offender.latestBooking()` to the `Offender` class instead of as an extension function? I suspect it's by far the most used.

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisoners/PrisonersResourceIntTest.kt
steverendell: I think the allBookings list being empty for aliases is correct - it is the case when loaded from the database too. The many to one is on the root offender id so only the root offender has bookings in the list. Thats why the getAllBookings() function is needed. And so i think the records returned from the dsls are correct too but i'll add assertions to the tests for that.

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisoners/PrisonersResourceIntTest.kt
steverendell: I did wonder about that, ok ill make aliases a separate builder

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
steverendell: I left it in because it is used quite widely in the tests in various ways to setup data etc and the volume of changes were getting out of hand. Maybe a follow-up PR

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/Offender.kt
steverendell: Ah, well it turned out there is no other way to define a prisoner with the current booking on an alias and an old booking on the root. Without this, the first booking always gets bookingSequence = 1.

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/integration/IntegrationTestBase.kt
steverendell: Good idea

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/Offender.kt
steverendell: Maybe its poorly named, its intended as the 'current' offender, used by the booking function to hook up its offender parent. When Ive reorganised things to use an alias dsl this may go

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
mikehalmamoj: OK cool. But I think it needs to go sooner rather than later if possible.

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/Offender.kt
mikehalmamoj: Well you possibly could by declaring the alias before the old booking... but yes I never liked that the active booking has to be the first one declared, it feels counterintuitive.

We could try and be clever by holding onto the list of bookings and juggling the sequences around to make the active booking no.1, and also adding some validation to prevent more than 1 active booking.

But that sounds horrible so in complicated scenarios declaring the bookingSequence for each booking is far more explicit. So I think you're right and we should keep it.

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
```diff
@@ -120,6 +99,11 @@ data class Offender(
   val middleNames: String
     get() = StringUtils.trimToNull(StringUtils.trimToEmpty(middleName) + " " + StringUtils.trimToEmpty(middleName2))
 
+  fun getAllBookings(): MutableList<OffenderBooking>? = rootOffender?.allBookings
+
+  fun latestBooking(): OffenderBooking =
+    getAllBookings()?.firstOrNull { it.active } ?: throw IllegalStateException("Offender has no active bookings")
```
andymarke, line 105: I think the isActive check isn't appropriate - I think latestBooking should do what it says on the tin.
do we need `activeBooking`

Also latestBooking should simply be find {it.bookingSequence == 1} no need for any ordering

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
```diff
@@ -76,34 +76,13 @@ data class Offender(
   @JoinColumn(name = "ROOT_OFFENDER_ID", updatable = false, insertable = false)
   var rootOffender: Offender? = null,
 
-  @Deprecated("Do not use; only some of the bookings may be associated with the root or any particular alias")
-  @OneToMany(mappedBy = "offender", cascade = [CascadeType.ALL], fetch = FetchType.LAZY, orphanRemoval = true)
-  val bookings: MutableList<OffenderBooking> = mutableListOf(),
+  // CAUTION: this list is only populated for the root offender; normally the function getAllBookings() below should be used instead
+  @OneToMany(mappedBy = "rootOffender", cascade = [CascadeType.ALL], fetch = FetchType.LAZY, orphanRemoval = true)
+  private val allBookings: MutableList<OffenderBooking> = mutableListOf(),
```
andymarke, line 81: so was it not possible to map this in JPA land for all offender aliases ? (I can imagine some sort cascading delete error ??)

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
```diff
@@ -76,37 +76,13 @@ data class Offender(
   @JoinColumn(name = "ROOT_OFFENDER_ID", updatable = false, insertable = false)
   var rootOffender: Offender? = null,
 
-  @Deprecated("Do not use; only some of the bookings may be associated with the root or any particular alias")
-  @OneToMany(mappedBy = "offender", cascade = [CascadeType.ALL], fetch = FetchType.LAZY, orphanRemoval = true)
-  val bookings: MutableList<OffenderBooking> = mutableListOf(),
-
-  @OneToMany(mappedBy = "rootOffender", fetch = FetchType.LAZY)
+  // CAUTION: this list is only populated for the root offender; normally the function getAllBookings() below should be used instead
+  @OneToMany(mappedBy = "rootOffender", cascade = [CascadeType.ALL], fetch = FetchType.LAZY, orphanRemoval = true)
```
mikehalmamoj, line 80: Is that CAUTION true? Doesn't seem to be a problem in the test I'm debugging...

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisoners/PrisonersResourceIntTest.kt
```diff
@@ -22,6 +24,173 @@ class PrisonersResourceIntTest : IntegrationTestBase() {
...
+          activeBooking.bookingId,
+          oldBooking.bookingId,
+          aliasBooking.bookingId,
+          oldAliasBooking.bookingId,
+        )
+    }
+  }
+
+  @Test
+  fun `DSL setup check - latest booking on alias`() {
```
mikehalmamoj, line 134: Can we have another test - where there is more than 1 root offender (with aliases and bookings on each)? I'm 99% sure it's fine, but that neat trick of passing the OffenderBuilder into the AliasBuilder means there's some state shared so would be good to make sure it scales up.

mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Offender.kt
mikehalmamoj: Sorry you're right - I was getting confused by Intelllij.

mikehalmamoj (approved):

PR #761, opened 2024-06-19
SDIT-1812 Return full booking start/end times

mikehalmamoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisonerprofile/PrisonerProfileIntTest.kt
```diff
@@ -68,14 +69,16 @@ class PrisonerProfileIntTest : IntegrationTestBase() {
     @Nested
     inner class HappyPath {
       lateinit var booking: OffenderBooking
-      val today = LocalDate.now()
+
+      // The DB column is a DATE type so truncates milliseconds, but bizarrely H2 uses half-up rounding so I have to emulate here or tests fail
+      val today = LocalDateTime.now().roundToNearestSecond()
```
petergphillips, line 74: might be better to call this `now` instead of `today`

petergphillips (approved):
LGTM

PR #771, opened 2024-07-02
SDIT-1816 fix booking end time

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisonperson/PrisonPersonService.kt
```diff
@@ -43,13 +43,13 @@ class PrisonPersonService(
...
+  // NOMIS truncates the time from booking end date, so try and get the accurate time from the last release movement
   private fun OffenderBooking.getReleaseTime(): LocalDateTime? =
     bookingEndDate?.let {
       externalMovements
-        .filter { it.movementDate == bookingEndDate?.toLocalDate() }
         .filter { it.movementType.code == "REL" }
         .maxByOrNull { it.movementTime }
         ?.movementTime
+        ?: bookingEndDate
     }
```
mikehalmamoj, line 54: This is the main change
* don't assume the release movement is on the same day as the booking end date
* if we don't find a release movement then fall back to the booking end date

petergphillips (approved):
LGTM

PR #774, opened 2024-07-09
SDIT-1887 Permanent location deactivation should succeed when already inactive

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/locations/LocationService.kt
```diff
@@ -117,7 +115,9 @@ class LocationService(
     val location = agencyInternalLocationRepository.findByIdOrNull(locationId)
       ?: throw NotFoundException("Location with id=$locationId does not exist")
 
-    if (!location.active) {
+    if (deactivateRequest.force) {
+      log.info("Force-deactivating already inactive location $location")
```
petergphillips, line 119: should this be in the `!location.active` if block?  Otherwise you'll get the message even if the location is active

steverendell: whoah good point! I wanted to see a message when force was true though so its the message that is wrong

petergphillips (approved):
LGTM

PR #783, opened 2024-07-19
SDIT-1902 endpoint to return filtered IDs for court case migration

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/CourtCaseRepository.kt
```diff
@@ -18,4 +21,31 @@ interface CourtCaseRepository : JpaRepository<CourtCase, Long> {
...
+    pageable: Pageable,
+  ): Page<Long>
+
+  @Query(
+    """
+      select
+        courtCase.id
+      from CourtCase courtCase 
+      order by courtCase.id asc
+    """,
```
lbennett-moj, line 48: updating created date after creation to allow filtering on ID testing

mikehalmamoj (approved):

PR #791, opened 2024-07-29
SDIT-1813 Return empty list for no bookings

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisonperson/PrisonPersonService.kt
```diff
@@ -14,12 +15,16 @@ import java.time.LocalDateTime
...
 ) {
-  fun getPhysicalAttributes(offenderNo: String) =
-    bookingRepository.findAllByOffenderNomsId(offenderNo)
-      .takeIf { it.isNotEmpty() }
-      ?.filterNot { it.physicalAttributes.isEmpty() }
-      ?.map {
+  fun getPhysicalAttributes(offenderNo: String): PrisonerPhysicalAttributesResponse {
+    if (!offenderRepository.existsByNomsId(offenderNo)) {
+      throw NotFoundException("No offender found for $offenderNo")
+    }
```
mikehalmamoj, line 23: We want to treat no booking the same as no physical attributes entered on bookings - this endpoint is just for a migration so zero is entirely expected. So we'll only return not found if the prisoner doesn't exist.

steverendell (approved):
lgtm

PR #801, opened 2024-08-06
SDIT-1826 Add telemetry when creating or updating physical attributes

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisonperson/PrisonPersonService.kt
```diff
@@ -66,7 +69,9 @@ class PrisonPersonService(
     }
 
   fun upsertPhysicalAttributes(offenderNo: String, request: UpsertPhysicalAttributesRequest): UpsertPhysicalAttributesResponse {
+    val telemetry = mutableMapOf("offenderNo" to offenderNo)
     val booking = bookingRepository.findLatestByOffenderNomsId(offenderNo)
+      ?.also { telemetry["bookingId"] = it.bookingId.toString() }
```
petergphillips, line 74: this looks a bit weird.  Couldn't you move the `val telemetry` to line 76 and then it could be an immutable map that also includes the `bookingId`?  If the booking is not found then an exception is thrown anyway

petergphillips: in fact looks like you could move it all to line 93?

mikehalmamoj: I thought I was going to be adding some error handling hence building up telemetry piecemeal. But don't need error handling so have moved.

petergphillips (dismissed):
LGTM

petergphillips (approved):
LGTM

PR #805, opened 2024-08-14
SDIT-1508 Detect duplicate visits

andymarke:
Check we can detect a duplicate visits efficiently by outputting telemetry. Next phase will be to reject the visit with a 409


petergphillips (approved):
LGTM

PR #808, opened 2024-08-21
SDIT-1829 Add get physical attributes for reconciliation endpoint

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisonperson/PrisonPersonService.kt
```diff
@@ -125,4 +145,14 @@ class PrisonPersonService(
     weightKilograms = value
     weightPounds = weightKilograms?.let { (it / 0.453592) }?.roundToInt()
   }
+
+  private fun List<OffenderPhysicalAttributes>.findLastModified(): OffenderPhysicalAttributes? =
+    takeIf { it.isNotEmpty() }
+      ?.maxBy {
```
karenmillermoj, line 151: just wondering why you would need maxBy ?

karenmillermoj (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisonperson/PrisonPersonService.kt
mikehalmamoj: There can be several height and weight records for an offender but any of them can be edited at any time. So we're only interested in the last one that was updated. We have the same logic in the sync and migration service.

PR #815, opened 2024-08-29
SDIT-1996 Add additional displayName fields for csip sync

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/CSIPAttendee.kt
```diff
@@ -46,10 +46,18 @@ class CSIPAttendee(
   @Column
   var auditModuleName: String? = null,
 
+  @ManyToOne(fetch = FetchType.LAZY)
+  @JoinColumn(name = "CREATE_USER_ID", insertable = false, updatable = false)
+  var createdByStaffUserAccount: StaffUserAccount? = null,
```
mikehalmamoj, line 51: Is this an n+1 problem? (same Q for modify)

mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/CSIPAttendee.kt
karenmillermoj: hmm -possibly I copied how OffenderAlert does it - but yes, it could possibly be an issue. Not sure whether to add an entitygraph or just check the sql when it gets run in anger and see if it's needed.

PR #816, opened 2024-08-30
SDIT-1830 Simplify reconciliation - just check latest height/weight from NOMIS

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisonperson/PrisonPersonService.kt
```diff
@@ -101,28 +101,17 @@ class PrisonPersonService(
...
+    val latestBooking = bookingRepository.findLatestByOffenderNomsId(offenderNo)
+      ?: throw NotFoundException("No bookings found for $offenderNo")
+
+    return latestBooking.physicalAttributes.minByOrNull { it.id.sequence }
+      ?.let {
+        PrisonPersonReconciliationResponse(
+          offenderNo = offenderNo,
+          height = it.getHeightInCentimetres(),
+          weight = it.getWeightInKilograms(),
+        )
```
mikehalmamoj, line 113: We were previously attempting to apply the same business rules as DPS when determining which is the "active" height and weight. Which is a bit silly because the reconciliation is meant to show that the same values are shown in DPS and NOMIS.

This does mean that the first reconciliation will highlight a lot of errors - mainly where heights/weight haven't been copied from old bookigns in NOMIS - but DPS are aware of this and plan to do a one-off update in NOMIS to align it with the DPS values.

karenmillermoj (approved):
lgtm

PR #817, opened 2024-09-04
SDIT-2026 Return existing visitId for 409 duplicate

andymarke:
Synchronisation service needs to know the existing visitId


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/data/Exceptions.kt
```diff
@@ -2,25 +2,25 @@ package uk.gov.justice.digital.hmpps.nomisprisonerapi.data
 
 import java.util.function.Supplier
 
-class BadDataException(message: String?, val error: BadRequestError? = null) : RuntimeException(message), Supplier<BadDataException> {
```
andymarke, line 5: just formatting

mikehalmamoj (approved):

PR #819, opened 2024-09-06
SDIT-2029 Fix migration bug

mikehalmamoj:
In NOMIS we found some inactive bookings without an end date - we were sending these null booking end date, but we should have sent them the last release date.


petergphillips (approved):
LGTM

PR #826, opened 2024-09-12
SDIT-1192 Do not set update immutable outcomeReason on visit visitor

andymarke:
outcomeReasonCode is immutable but the shadow outcomeReason is read-only


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/visits/VisitServiceTest.kt
```diff
@@ -115,7 +115,8 @@ internal class VisitServiceTest {
     bookingId = OFFENDER_BOOKING_ID,
     offender = defaultOffender,
     bookingBeginDate = LocalDateTime.now(),
-  ).apply { // add circular reference
+  ).apply {
+    // add circular reference
```
andymarke, line 119: ktlint

mikehalmamoj (approved):

PR #828, opened 2024-09-13
SDIT-1996 Add displayName to CSIP Factor

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/csip/Transformer.kt
```diff
@@ -78,6 +78,19 @@ private fun CSIPReport.toSCSResponse() =
...
   )
 
+fun CSIPFactor.toFactorResponse() =
+  CSIPFactorResponse(
+    id = id,
+    type = type.toCodeDescription(),
+    comment = comment,
+    createDateTime = createDatetime,
+    createdBy = createUsername,
+    createdByDisplayName = createdByStaffUserAccount.asDisplayName(),
```
andymarke, line 88: so is createdByStaffUserAccount guaranteed *never* to be null ? is there a foreign key constraint or whatever?

andymarke (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/csip/Transformer.kt
karenmillermoj: yes - it works for null.

PR #831, opened 2024-09-16

andymarke:

TODO Need to also move visit creation to DSL

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderVisitBalance.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderVisitBalance.kt
@@ -45,5 +45,5 @@ data class OffenderVisitBalance(
   override fun hashCode(): Int = javaClass.hashCode()
 
   @Override
-  override fun toString(): String = this::class.simpleName + "(offenderBookingId = $offenderBookingId )"
+  override fun toString(): String = "${this::class.simpleName} (offenderBookingId = $offenderBookingId)"
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderVisitBalance.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderVisitBalance.kt
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa
 
 import jakarta.persistence.Column
+import jakarta.persistence.Convert
 import jakarta.persistence.Entity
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderVisitBalance.kt
```diff
@@ -29,6 +31,7 @@ data class OffenderVisitBalance(
   var remainingPrivilegedVisitOrders: Int? = null,
 
   @Column(name = "VISIT_ALLOWANCE_INDICATOR")
+  @Convert(converter = YesNoConverter::class)
   val visitAllowanceIndicator: Boolean? = false,
 ) {
```
andymarke, line 36: was missing, but we only insert via tests so not a live bug

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderVisitBalance.kt
```diff
@@ -42,7 +45,5 @@ data class OffenderVisitBalance(
   override fun hashCode(): Int = javaClass.hashCode()
 
   @Override
-  override fun toString(): String {
-    return this::class.simpleName + "(offenderBookingId = $offenderBookingId )"
-  }
+  override fun toString(): String = this::class.simpleName + "(offenderBookingId = $offenderBookingId )"
```
petergphillips, line 48:
```suggestion
  override fun toString(): String = "${this::class.simpleName} (offenderBookingId = $offenderBookingId )"
```

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/Contact.kt
```diff
@@ -0,0 +1,62 @@
...
+interface OffenderContactPersonDsl
+
+@Component
+class OffenderContactPersonBuilderRepository(
+  private val offenderContactPersonRepository: OffenderContactPersonRepository,
+  private val relationshipTypeRepository: ReferenceCodeRepository<RelationshipType>,
+  private val contactTypeRepository: ReferenceCodeRepository<ContactType>,
+) {
+  fun save(contact: OffenderContactPerson): OffenderContactPerson = offenderContactPersonRepository.save(contact)
+  fun lookupRelationshipType(code: String): RelationshipType = relationshipTypeRepository.findByIdOrNull(Pk(RelationshipType.RELATIONSHIP, code))!!
```
petergphillips, line 27: would be better to call `getReferenceById` and then you wouldn't need the `!!`.  It would also throw a `EntityNotFoundException` rather than a null pointer too

petergphillips (approved):
LGTM

PR #832, opened 2024-09-16

andymarke:

TODO Need add phone/address to Person


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/LegacyVisitBuilder.kt
```diff
@@ -10,15 +10,15 @@ import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.VisitType
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.VisitVisitor
 import java.time.LocalDateTime
 
-class VisitBuilder(
+class LegacyVisitBuilder(
```
petergphillips, line 13: would be good to understand why this is `Legacy` - should it therefore be deprecated?  Should it also then link to the replacement somehow?

petergphillips (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/LegacyVisitBuilder.kt
andymarke: They will be replaced very soon, but they clash with new class names that will replace them

PR #834, opened 2024-09-16

andymarke:

TODO Fix up unit tests


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/AddressPhone.kt
```diff
@@ -0,0 +1,31 @@
...
+
+@NomisDataDslMarker
+interface AddressPhoneDsl
+
+@Component
+class AddressPhoneBuilderFactory {
+  fun builder() = AddressPhoneBuilderRepositoryBuilder()
+}
+
+class AddressPhoneBuilderRepositoryBuilder : AddressPhoneDsl {
```
mikehalmamoj, line 18: Can phone numbers also be attached directly to a person (rather than an address)?

mikehalmamoj (approved):

PR #836, opened 2024-09-17
SDIT-2019 Allow null profile codes

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisonperson/profiledetails/ProfileDetailsService.kt
```diff
@@ -28,11 +28,10 @@ class ProfileDetailsService(
           endDateTime = booking.getReleaseTimer(),
           latestBooking = booking.bookingSequence == 1,
           profileDetails = booking.profiles.first().profileDetails
-            .filter { pd -> pd.profileCodeId != null }
```
mikehalmamoj, line 31: Records with null profile codes get created for new offenders but also exist where the user has cleared the profile code. Don't filter them out here, that's up to the caller of the API.

andymarke (approved):
lgtm

PR #842, opened 2024-09-18
SDIT-2044 A few more fields to PERSON

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/contactperson/ContactPersonResource.kt
```diff
@@ -101,4 +102,14 @@ data class ContactPerson(
...
+  @Schema(description = "Date the person dies")
+  val deceasedDate: LocalDate?,
+  @Schema(description = "True if a staff member")
+  val isStaff: Boolean?,
+  @Schema(description = "No longer used in NOMIS since 2018")
+  val isRemitter: Boolean?,
+  @Schema(description = "No longer used in NOMIS since 2019")
+  val keepBiometrics: Boolean,
+  @Schema(description = "Audit data associated with the records")
+  val audit: NomisAudit,
```
andymarke, line 114: No idea if this will be needed yet (same as above ) but added as a discussion point

mikehalmamoj (approved):

PR #850, opened 2024-09-24
SDIT-2044 A few more fields to PERSON ADDRESS

andymarke:
Primary, Mail, PAF


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/contactperson/ContactPersonResource.kt
```diff
@@ -156,4 +156,12 @@ data class Address(
   val country: CodeDescription?,
   @Schema(description = "List of phone numbers for the address")
   val phoneNumbers: List<PhoneNumber>,
+  @Schema(description = "true if address validated by Post Office Address file??")
+  val validatedPAF: Boolean,
```
andymarke, line 160: not happy with this name, so trying to find out what this actually means

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Address.kt
```diff
@@ -40,7 +42,13 @@ abstract class Address(
   open val startDate: LocalDate = LocalDate.now(),
 
   @Column(name = "NO_FIXED_ADDRESS_FLAG")
-  open val noFixedAddressFlag: String = "N",
+  @Convert(converter = YesNoConverter::class)
```
andymarke, line 45: renamed and switched to boolean

mikehalmamoj (approved):

PR #854, opened 2024-09-25
SDIT-2053 Switch to mutating visit order list rather than recreating

andymarke:
This might mitigate the locking on visit order visitor table

Not convinced this will fix the issue, but should reduce the occurrences since no DELETE will happen if no visitors removed.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/visits/VisitService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/visits/VisitService.kt
@@ -224,19 +224,25 @@ class VisitService(
     visit.visitors.removeAll(visitorsToRemove)
     visit.visitors.addAll(visitorsToAdd)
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/visits/VisitService.kt
```diff
@@ -224,19 +224,25 @@ class VisitService(
...
-        visit.visitors
-          .filter { it.person != null }.mapIndexed { index, visitor ->
-            VisitOrderVisitor(
-              person = visitor.person!!,
-              groupLeader = index == 0,
-              visitOrder = visitOrder,
-            )
-          },
-      )
+    visit.visitOrder?.also { visitOrder ->
```
andymarke, line 227: essentially the same type of code we do for visitors; just add/remove visitors to the order that only of they have actually been added or removed. Previous all rows were deleted and re-added

steverendell (approved):
lgtm

PR #855, opened 2024-09-25
SDIT-2053 Wait for visit order visitors lock to clear

andymarke:
Will wait 10 seconds then will throw exception, else this would wait indefinitely

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/VisitOrderVisitorRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/VisitOrderVisitorRepository.kt
@@ -1,0 +1,21 @@
+package uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.repository
+
+import jakarta.persistence.LockModeType
+import jakarta.persistence.QueryHint
+import org.springframework.data.jpa.repository.Lock
+import org.springframework.data.jpa.repository.QueryHints
+import org.springframework.data.repository.CrudRepository
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.VisitOrderVisitor
+
+@Repository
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/VisitOrderVisitorRepository.kt
```diff
@@ -0,0 +1,21 @@
...
+@Repository
+interface VisitOrderVisitorRepository : CrudRepository<VisitOrderVisitor, Long> {
+  /**
+   * Will do a "select ... for update" so will wait for any locks to clear
+   * This means if the NOMIS screen has locked these rows it will hang for 10 seconds and then throw exception
+   * and the whole process will be retried until the lock disappears.
+   */
+  @Lock(LockModeType.PESSIMISTIC_WRITE)
+  @QueryHints(value = [QueryHint(name = "jakarta.persistence.lock.timeout", value = "10000")])
+  fun findAllByIdIn(visitorIds: List<Long>): List<VisitOrderVisitor>
```
steverendell, line 20: Did you not want to wait any longer than 10s? As it is a balance between protecting the DB vs maximising the chance that if the user goes back into the Nomis screen they see the change, i get it if you go more for DB

steverendell (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/visits/VisitService.kt
```diff
@@ -236,6 +238,12 @@ class VisitService(
           groupLeader = false,
         )
       }
+
+      if (voVisitorsToRemove.isNotEmpty()) {
+        // wait for any locks in NOMIS
```
mikehalmamoj, line 243: Better to have the word "lock" in the method name rather than this comment? maybe `lockAllByIdIn`.

andymarke: Hmm, doesn't it have to be the Spring Data name given there is no HQL?

PR #858, opened 2024-09-25
SDIT-2058 Some older casenotes amendments fail to be identified

steverendell:
the date format has changed over time


andymarke (approved):
lgtm

PR #860, opened 2024-09-25
SDIT-2044 Add person employment and employer

andymarke:
Employer in NOMIS is corporate so mapped this as well, though it is likely at most we need to just send back the corporate id to DPS


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/NomisData.kt
```diff
@@ -500,6 +535,19 @@ interface NomisDataDsl {
     keepBiometrics: Boolean = false,
     dsl: PersonDsl.() -> Unit = {},
   ): Person
+
+  fun corporate(
```
mikehalmamoj, line 539: There's no DSL marker annotation here - but there is in the implementation class. Not sure which is right or if it matters?

mikehalmamoj (approved):

PR #861, opened 2024-09-25
SDIT-2044 Fix unnecessary DSL Markers

andymarke:


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/NomisData.kt
```diff
@@ -536,6 +525,7 @@ interface NomisDataDsl {
     dsl: PersonDsl.() -> Unit = {},
   ): Person
 
+  @CorporateDslMarker
```
andymarke, line 528: this was missing

mikehalmamoj (approved):

PR #867, opened 2024-09-27
SDIT-2044 Added person global restrictions

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderContactPerson.kt
```diff
@@ -94,7 +94,7 @@ data class OffenderContactPerson(
   val expiryDate: LocalDate? = null,
 
   @OneToMany(mappedBy = "contactPerson", cascade = [CascadeType.ALL], fetch = LAZY)
-  val restrictions: MutableList<OffenderPersonRestricts> = mutableListOf(),
+  val restrictions: MutableList<OffenderPersonRestrict> = mutableListOf(),
```
andymarke, line 97: removed plural for consitency

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/contactperson/ContactPersonService.kt
```diff
@@ -106,6 +106,16 @@ class ContactPersonService(private val personRepository: PersonRepository) {
           },
         )
       },
+      restrictions = it.restrictions.map { restriction ->
+        ContactRestriction(
+          id = restriction.id,
+          type = restriction.restrictionType.toCodeDescription(),
```
mikehalmamoj, line 112: Is description right here? Feels like a code...

mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/contactperson/ContactPersonService.kt
andymarke: type? so type will be type.code = "BAN" type.description = "Banned"

so "type" refers to the "restriction" "type"

mikehalmamoj (approved):

PR #886, opened 2024-10-08
SDIT-1981 Add dpsId to each csip request and remove setting of createusername

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/csip/UpsertCSIPRequest.kt
```diff
@@ -41,9 +40,6 @@ data class UpsertCSIPRequest(
   @Schema(description = "If assaulted, the staff member name")
   val staffAssaultedName: String? = null,
 
-  @Schema(description = "Audit detail info")
```
karenmillermoj, line 44: Audit details not required in any request - removed from all

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/csip/CSIPService.kt
```diff
@@ -132,6 +134,7 @@ class CSIPService(
       ?: log.info("CSIP deletion request for: $csipId ignored. CSIP does not exist")
   }
 
+  @Audit
```
andymarke, line 137: Are you sure we still need that annotation ? I think it was originally needed for the module_name thing but is now done automatically

karenmillermoj: I wasn't sure tbh - I will remove (I think I copied one of the other places that had it set) - I was gonna speak to you about it but then forgot. - will remove in next PR

PR #893, opened 2024-10-10
SDIT-2089 Rename DTO classes to avoid clashes

andymarke:
To avoid DTO names that match other API endpoints


petergphillips (approved):
LGTM

PR #898, opened 2024-10-15
SDIT-1981 Insert all csip objects and identify new plans in response

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/csip/CSIPService.kt
```diff
@@ -309,23 +417,37 @@ class CSIPService(
...
-            this.factors[index].id,
-            requestFactor.dpsId,
-          ),
-        )
-      }
-    }
+fun CSIPReport.identifyNewComponents(request: UpsertCSIPRequest) =
+  identifyNewFactors(request) +
+    identifyNewPlans(request)
+// + identifyNewReviews(request)
```
karenmillermoj, line 423: still to do - will add these together and use interface to avoid duplication

petergphillips (approved):
LGTM

PR #907, opened 2024-10-24
SDIT-2138 Handle start dates from activity pay rate updates

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/PayRatesService.kt
```diff
@@ -63,116 +63,97 @@ class PayRatesService(
...
+   * To handle the delete problem we expire rates in NOMIS that are no longer in the DPS requested rates.
    */
-  fun buildNewPayRates(requestedPayRates: List<PayRateRequest>, existingActivity: CourseActivity, previousActivityEndDate: LocalDate?): MutableList<CourseActivityPayRate> {
-    val newPayRates = mutableListOf<CourseActivityPayRate>()
-    val existingPayRates = existingActivity.payRates
-    val tomorrow = LocalDate.now().plusDays(1)
-
-    requestedPayRates.forEach { requestedPayRate ->
-      val existingPayRate = existingPayRates.findExistingPayRate(requestedPayRate)
+  fun buildNewPayRates(requestedPayRates: List<PayRateRequest>, existingActivity: CourseActivity): List<CourseActivityPayRate> {
```
mikehalmamoj, line 74: The main change is to rewrite the algorithm to handle the new DPS model. 

It would be great if their model matched NOMIS then we could perform a straight mapping. Unfortunately deleting pay rate history from NOMIS is not an option because people can be paid in the future so we need to keep the old rates.

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/ActivityResourceIntTest.kt
```diff
@@ -948,44 +982,6 @@ class ActivityResourceIntTest : IntegrationTestBase() {
         }
       }
 
-      @Test
-      fun `should retain the end date if pay rate start date is moved back at same time as rate is changed`() {
```
mikehalmamoj, line 952: Removed because this test doesn't make sense - it's not possible to move the start date back.

steverendell (approved):
lgtm

PR #909, opened 2024-10-25
SDIT-2143 Add CSIP Reconciliation

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/csip/CSIPResource.kt
```diff
@@ -84,11 +84,11 @@ class CSIPResource(private val csipService: CSIPService) {
   fun upsertCSIP(@RequestBody @Valid request: UpsertCSIPRequest) = csipService.upsertCSIP(request)
 
   @PreAuthorize("hasRole('ROLE_NOMIS_CSIP')")
-  @GetMapping("/prisoners/{offenderNo}/csip/to-migrate")
+  @GetMapping("/prisoners/{offenderNo}/csip/reconciliation")
```
karenmillermoj, line 87: Now used for reconciliation as migration now uses different endpoint

andymarke (approved):

PR #915, opened 2024-10-28
SDIT-2148 Return Audit information for person addresses and email address

andymarke:


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/contactperson/ContactPersonResourceIntTest.kt
```diff
@@ -173,9 +173,6 @@ class ContactPersonResourceIntTest : IntegrationTestBase() {
       @BeforeEach
       fun setUp() {
         nomisDataBuilder.build {
-          staff(firstName = "KOFE", lastName = "ADDY") {
```
andymarke, line 176: no needed since it is the outer class setup()

petergphillips (approved):
LGTM

PR #919, opened 2024-10-29
SDIT-2149 Contact restriction enter by username

andymarke:
Return general username for entered restriction so that DPS can use username rather than NOMIS staffId

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationService.kt
@@ -12,6 +12,7 @@ import uk.gov.justice.digital.hmpps.nomisprisonerapi.data.BadDataException
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.data.CodeDescription
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.data.NotFoundException
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.data.toCodeDescription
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationService.kt
```diff
@@ -1344,9 +1334,6 @@ fun uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.Staff.toStaff() =
     username = accounts.usernamePreferringGeneralAccount(),
   )
 
-private fun List<StaffUserAccount>.usernamePreferringGeneralAccount() =
```
andymarke, line 1347: actually change here - move this to shared file

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helpers/NomisStaff.kt
```diff
@@ -0,0 +1,10 @@
+package uk.gov.justice.digital.hmpps.nomisprisonerapi.helpers
+
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.Staff
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.StaffUserAccount
+
+fun List<StaffUserAccount>.usernamePreferringGeneralAccount() =
+  this.maxByOrNull { it.type }?.username ?: "unknown"
```
andymarke, line 7: GENERAL will be used over ADMIN, for multiple GENERAL we will get one at random

mikehalmamoj (approved):

PR #929, opened 2024-11-06
SDIT-2191 Race condition with multiple location events

steverendell:
Ensure multiple events for the same location are processed sequentially.


andymarke (approved):
lgtm

PR #932, opened 2024-11-11
SDIT-2192 Add endpoint to retrieve splash screen config

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/core/SplashScreenResource.kt
```diff
@@ -0,0 +1,81 @@
...
+import org.springframework.web.bind.annotation.RestController
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.config.ErrorResponse
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.data.CodeDescription
+
+@RestController
+@Validated
+@RequestMapping(produces = [MediaType.APPLICATION_JSON_VALUE])
+class SplashScreenResource(private val service: SplashScreenService) {
+
+  @PreAuthorize("hasRole('ROLE_NOMIS_SCREEN_ACCESS')")
```
karenmillermoj, line 22: Wasn't sure of a role name to use (there may be a  more general one for admin?)

src/main/resources/db/migration/V1_1101__SPLASH_SCREEN_FUNCS.sql
```diff
@@ -0,0 +1,18 @@
+create table SPLASH_SCREEN_FUNCS
```
karenmillermoj, line 1: This table is not used in Nomis but the table I am interested in depends on it.
The table allows pl/sql functions to be added - but currently this is not done

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/core/SplashScreenResource.kt
andymarke: Difficult ones for these end points that cross domains. So maybe stick with this

andymarke (approved):
LGTM

PR #935, opened 2024-11-12

andymarke:
Can optional supply an Id that will be used instead of sequence


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Person.kt
```diff
@@ -25,8 +29,17 @@ import java.time.LocalDate
 data class Person(
   @Id
   @Column(name = "PERSON_ID")
-  @SequenceGenerator(name = "PERSON_ID", sequenceName = "PERSON_ID", allocationSize = 1)
   @GeneratedValue(generator = "PERSON_ID")
+  // TODO - work out how to use IdGeneratorType instead of deprecated GenericGenerator
```
andymarke, line 33: GenericGenerator is deprecated - after 4 hours I gave up trying to use the new annotation.
I couldn't work out how to get parameters from annotation into a super class SequenceStyleGenerator

petergphillips (approved):
LGTM

PR #937, opened 2024-11-12

andymarke:
Replace deprecated class with SequenceOrUseId


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/SequenceGeneratorOrUseId.kt
```diff
@@ -0,0 +1,37 @@
...
+    appliedParams[INITIAL_PARAM] = config.startWith
+    appliedParams[INCREMENT_PARAM] = config.incrementBy
+    appliedParams[SEQUENCE_PARAM] = config.name
+
+    super.configure(type, appliedParams, serviceRegistry)
+  }
+
+  @Throws(HibernateException::class)
+  override fun generate(session: SharedSessionContractImplementor, entity: Any): Any {
+    val id = session.getEntityPersister(null, entity).getIdentifier(entity, session)
```
andymarke, line 34: slight improvement - look for id and compare with 0L which we have for all NOMIS tables - previously this hardcode the Person entity

PR #939, opened 2024-11-13

andymarke:
Fix hardcoded ID


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/contactperson/ContactPersonService.kt
```diff
@@ -183,7 +183,7 @@ class ContactPersonService(
 
     request.let {
       Person(
-        id = 98765443,
+        id = it.personId ?: 0,
```
lbennett-moj, line 186: 😺

lbennett-moj (approved):

PR #940, opened 2024-11-13

andymarke:
Minor changes to be consistent with NOMIS

NOMIS Person search only works for uppercase names


petergphillips (approved):
LGTM

PR #941, opened 2024-11-14
SDIT-2211 JPA models for OFFENDER_IMAGES table

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderImage.kt
```diff
@@ -0,0 +1,98 @@
...
+      ), JoinColumnOrFormula(column = JoinColumn(name = "IMAGE_SOURCE_CODE", referencedColumnName = "code")),
+    ],
+  )
+  val imageSource: ImageSource,
+
+  @Column(name = "ORIENTATION_TYPE")
+  val orientationTypeCode: String,
+
+  @Column(name = "IMAGE_VIEW_TYPE")
+  val imageViewTypeCode: String,
```
mikehalmamoj, line 76: These columns appear to reference different reference codes for different image object types (bookings or identifying marks) hence we need some inheritance so we can model them differently.

andymarke (approved):
LGTM

PR #958, opened 2024-11-27
SDIT-2263 update court charge  with additional CEC update

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingService.kt
```diff
@@ -399,6 +399,21 @@ class SentencingService(
...
+  private fun refreshCourtEventCharges(
+    offenderCharge: OffenderCharge,
+    case: CourtCase,
+  ) {
+    case.courtEvents.flatMap { it.courtEventCharges }.filter { it.id.offenderCharge == offenderCharge }
+      .forEach { courtEventCharge ->
+        courtEventCharge.offenceDate = offenderCharge.offenceDate
+        courtEventCharge.offenceEndDate = offenderCharge.offenceEndDate
+        courtEventCharge.mostSeriousFlag = offenderCharge.mostSeriousFlag
+        courtEventCharge.resultCode1 = offenderCharge.resultCode1
```
andymarke, line 412: Actually are you sure about the result code? Wouldn't that mean if the outcome on the final appearance was "Imprisonment" wouldn't it then update the first appearance result to "imprisonment" .
It doesn't "feel" right. It does for offenceDate etc but not the result

andymarke (approved):
LGTM but not sure about resultCode1

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingService.kt
lbennett-moj: will raise it with DPS as I don't think they have the ability to have different charge data for each appearance

PR #961, opened 2024-11-29
SDIT-2248 Revised casenotes merge strategy

steverendell:
Fix created date field returned from get casenotes


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/casenotes/CaseNotesService.kt
```diff
@@ -124,8 +122,8 @@ class CaseNotesService(
     caseNoteText = parseMainText(caseNoteText),
     amendments = parseAmendments(this),
     noteSourceCode = noteSourceCode,
-    createdDatetime = LocalDateTime.of(dateCreation, timeCreation?.toLocalTime() ?: LocalTime.MIDNIGHT),
-    createdUsername = createdUserId!!,
+    createdDatetime = createdDatetime,
+    createdUsername = createdUserId,
```
steverendell, line 126: We now need the DB timestamp, and dateCreation / timeCreation value is already available in the creationDateTime dto field

mikehalmamoj (approved):

PR #977, opened 2024-12-13
SDIT-1986 Change date to dateTime for incident history and pick up questionnaire updates

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/QuestionnaireAnswer.kt
```diff
@@ -51,6 +52,9 @@ data class QuestionnaireAnswer(
   @Convert(converter = YesNoConverter::class)
   val dateRequired: Boolean = false,
 
+  @Column(name = "EXPIRY_DATE")
```
karenmillermoj, line 55: The changes to questionnaire tables are just here for completeness, we don't return this data (initially we thought we might) but added to db mappings to ensure that they now match the updated ones in prison-api

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incidents/IncidentResource.kt
```diff
@@ -478,7 +478,7 @@ data class History(
   @Schema(description = "Questions asked for the questionnaire")
   val questions: List<HistoryQuestion>,
   @Schema(description = "When the questionnaire was changed")
-  val incidentChangeDate: LocalDate,
+  val incidentChangeDateTime: LocalDateTime,
   @Schema(description = "Who changed the questionnaire")
```
andymarke, line 482: Are you happy that this is a breaking change? You will break the migration service?

andymarke (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incidents/IncidentResource.kt
karenmillermoj: I think so - I will make sure both are deployed together

PR #986, opened 2024-12-18
SDIT-2301 Delete endpoints for all contact entities

andymarke:
Will return 204 for delete or if the entity does not exist. The exception is identifier that is a composite key so 404 for not found makes more sense if person is not found.

400 when the combination of child / parents entities do not match since this indicates a client coding issue.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/contactperson/ContactPersonService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/contactperson/ContactPersonService.kt
@@ -310,6 +310,13 @@ class ContactPersonService(
     }
   }
 
+  fun deletePersonContact(personId: Long, contactId: Long) {
+    contactRepository.findByIdOrNull(contactId)?.also {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/contactperson/ContactPersonService.kt
```diff
@@ -432,20 +471,10 @@ class ContactPersonService(
       }
     }
   }
-  fun createPersonContactRestriction(
```
andymarke, line 435: just moved to be next to the update

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/contactperson/ContactPersonService.kt
```diff
@@ -477,6 +506,28 @@ class ContactPersonService(
     }
   }
 
+  fun deletePersonRestriction(personId: Long, personRestrictionId: Long) {
```
andymarke, line 509: moved

steverendell (approved):
lgtm

PR #991, opened 2024-12-20
SDIT-2365 tweak to Court event charge endpoint

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingResource.kt
```diff
@@ -1124,7 +1124,7 @@ class SentencingResource(private val sentencingService: SentencingService) {
   ): OffenderChargeResponse = sentencingService.getOffenderCharge(offenderChargeId, offenderNo)
 
   @PreAuthorize("hasRole('ROLE_NOMIS_SENTENCING')")
-  @GetMapping("/prisoners/{offenderNo}/sentencing/court-cases/{caseId}/court-appearances/{eventId}/charges/{chargeId}")
```
lbennett-moj, line 1127: case isn't necessary and requires further call during synchronisation

PR #993, opened 2024-12-30
SDIT-2369 Truncate Alert comment text to 4000

andymarke:
Improve the truncate function to deal with multi-byte chars at the end of String.

Move to Util file as well so used for adjudications and alerts

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationService.kt
@@ -1,6 +1,5 @@
 package uk.gov.justice.digital.hmpps.nomisprisonerapi.adjudications
 
-import com.google.common.base.Utf8
 import com.microsoft.applicationinsights.TelemetryClient
 import org.slf4j.LoggerFactory
 import org.springframework.data.domain.Page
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationService.kt
```diff
@@ -1360,9 +1360,3 @@ fun AdjudicationHearingResultAward.toAward(isConsecutiveAward: Boolean = false):
       null
     },
   )
-
```
andymarke, line 1363: move to own file

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helpers/StringUtils.kt
```diff
@@ -0,0 +1,20 @@
+package uk.gov.justice.digital.hmpps.nomisprisonerapi.helpers
+
+import com.google.common.base.Utf8
+
```
andymarke, line 4: rewrote this so it only truncates enough so it fits, else Strings with lots of quad-byte characters at beginning of String were being truncate too much.,
Also deal with truncating half-way through a multi-byte character by accident

src/main/resources/db/migration/V1_53__OFFENDER_ALERTS.sql
```diff
@@ -11,7 +11,7 @@ CREATE TABLE "OFFENDER_ALERTS"
   "ALERT_STATUS"                  VARCHAR2(12 CHAR)                 NOT NULL ,
   "VERIFIED_FLAG"                 VARCHAR2(1 CHAR) DEFAULT 'N'      NOT NULL ,
   "EXPIRY_DATE"                   DATE,
-  "COMMENT_TEXT"                  VARCHAR2(1000 CHAR),
+  "COMMENT_TEXT"                  VARCHAR2(4000 CHAR),
   "CASELOAD_ID"                   VARCHAR2(6 CHAR),
```
andymarke, line 15: 4000 in live - no idea how this was ever 1000 in the test schema

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helpers/TruncateToUtf8LengthTest.kt
```diff
@@ -25,7 +25,12 @@ class TruncateToUtf8LengthTest {
   }
 
   @Test
-  fun `will truncate when too big with huge byte characters`() {
+  fun `will truncate when too big with huge byte characters at beginning`() {
     assertThat("😀123456789".truncateToUtf8Length(10)).isEqualTo("😀123456")
   }
+
+  @Test
```
andymarke, line 32: this test previously would fail, since too many characters would have been cut

petergphillips (approved):
LGTM

PR #994, opened 2025-01-06
SDIT-2165 Return total paid to check during attendance reconciliation

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderCourseAttendanceRepository.kt
```diff
@@ -18,15 +18,16 @@ interface OffenderCourseAttendanceRepository : JpaRepository<OffenderCourseAtten
...
+      join OffenderTransactionDetail otd on oca.transaction = otd.transaction and oca.eventId = otd.courseAttendance.eventId
       where oca.prison.id = :prisonId
       and oca.eventDate = :date
       and oca.pay = true
       group by oca.offenderBooking.bookingId
       order by oca.offenderBooking.bookingId
     """,
   )
-  fun findBookingPaidAttendanceCountsByPrisonAndDate(prisonId: String, date: LocalDate): List<BookingCount>
+  fun findBookingPaidAttendanceCountsByPrisonAndDate(prisonId: String, date: LocalDate): List<BookingCountWithPay>
```
mikehalmamoj, line 32: This is the main change - we now return the amount paid as well as the number of attendances recorded so that we can compare the amount pai with DPS during attendance reconciliation.

lbennett-moj (approved):
lgtm

PR #995, opened 2025-01-06
SDIT-2384 GET /bookings/ids/latest-from-id

andymarke:
New endpoint initial to be used by sentence adjustments reconcillation


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderBookingRepository.kt
```diff
@@ -30,4 +30,40 @@ interface OffenderBookingRepository :
...
+      from OFFENDER_BOOKINGS b join OFFENDERS o on b.offender_id = o.offender_id
+      where OFFENDER_BOOK_ID > :bookingId
+        and b.BOOKING_SEQ = 1
+        and b.ACTIVE_FLAG = 'Y'
+        and rownum <= :pageSize
+      order by OFFENDER_BOOK_ID
+    """,
+    nativeQuery = true,
+  )
+  fun findAllLatestActiveIdsFromId(bookingId: Long, pageSize: Int): List<BookingWithIds>
```
andymarke, line 63: two separate queries since I have observed trying to combine in to one with a parameter seems to be slowed when you don't need to filter by ACTIVE_FLAG

petergphillips (commented):
LGTM

petergphillips (approved):
LGTM

PR #999, opened 2025-01-07
SDIT-2162 Include activity start date in allocation response

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/api/GetAllocationResponse.kt
```diff
@@ -46,4 +46,7 @@ data class GetAllocationResponse(
 
   @Schema(description = "Rules for creating schedules - days and times")
   val scheduleRules: List<ScheduleRulesResponse> = listOf(),
+
+  @Schema(description = "The start date of the activity", example = "2023-03-12")
+  val activityStartDate: LocalDate,
```
mikehalmamoj, line 51: We're going to start allowing migration of activities that start in the future. When this happens we'll need to make sure that the allocation doesn't start before the activity or this will break DPS - hence getting activity start date in the allocation response.

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/GetAllocationResourceIntTest.kt
```diff
@@ -988,25 +988,29 @@ class GetAllocationResourceIntTest : IntegrationTestBase() {
             }
           }
           offender {
-            booking(agencyLocationId = "LEI") { // wrong prison
+            // wrong prison
+            booking(agencyLocationId = "LEI") {
```
mikehalmamoj, line 992: ktlint

petergphillips (approved):
LGTM

PR #1014, opened 2025-01-13
SDIT-2409 Core corporate details for GET /corporates/{corporateId}

andymarke:
PR comments


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Caseload.kt
```diff
@@ -15,7 +15,7 @@ class Caseload(
   var id: String,
   @Column(name = "DESCRIPTION", nullable = false)
   val description: String,
-) : NomisAuditableEntity() {
+) {
```
andymarke, line 18: we never insert these right now

mikehalmamoj (approved):

PR #1035, opened 2025-01-15
SDIT-2419 Increase heap size

mikehalmamoj:


helm_deploy/hmpps-nomis-prisoner-api/values.yaml
```diff
@@ -25,7 +25,6 @@ generic-service:
 
   # Environment variables to load into the deployment
   env:
-    JAVA_OPTS: "-Xmx512m"
```
mikehalmamoj, line 28: This does nothing, so remove to avoid confusion.

helm_deploy/hmpps-nomis-prisoner-api/values.yaml
```diff
@@ -44,7 +43,7 @@ generic-service:
       memory: 640Mi
     limits:
       cpu: 2000m
-      memory: 1240Mi
```
mikehalmamoj, line 47: The default heap size is 1/4 of the maximum memory available to the pod. This application's heap was reaching up to 370m which explains why we saw OOM errors.

We could explicitly set the size of all JVM memory components including the heap as explained in [this article](https://cloud.theodo.com/en/blog/jvm-oom).  But 1/4 seems fairly sensible to me and it's the default for a reason.

petergphillips (approved):
LGTM

PR #1036, opened 2025-01-17
SDIT-2396: ✨ Add start, end and latest to booking info

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/coreperson/CorePersonService.kt
```diff
@@ -59,6 +60,9 @@ class CorePersonService(
           b.profileDetails.filter { pd -> pd.id.profileType.type == "NAT" }.map { n ->
             OffenderNationality(
               bookingId = b.bookingId,
+              startDateTime = b.bookingBeginDate,
```
petergphillips, line 63: logic copied from prison person

mikehalmamoj (approved):

PR #1038, opened 2025-01-17
SDIT-2396: ✨ Ignore profile details where the value isn't set

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/coreperson/CorePersonResource.kt
```diff
@@ -230,7 +230,7 @@ data class OffenderNationality(
   @Schema(description = "The booking's unique identifier", example = "1234567")
   val bookingId: Long,
   @Schema(description = "The value of the profile info")
-  val nationality: CodeDescription?,
+  val nationality: CodeDescription,
```
petergphillips, line 233: we now ignore rows where there is no nationality etc. set

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/coreperson/CorePersonService.kt
```diff
@@ -57,59 +57,69 @@ class CorePersonService(
...
-          }
+          b.profileDetails.filter { it.id.profileType.type == "NAT" }
+            .filter { it.profileCodeId != null }
+            .map { n ->
+              OffenderNationality(
+                bookingId = b.bookingId,
+                startDateTime = b.bookingBeginDate,
+                endDateTime = b.getReleaseTime(),
+                latestBooking = b.bookingSequence == 1,
+                nationality = n.profileCode!!.toCodeDescription(),
```
petergphillips, line 68: bit naff the `!!` here, but we've already filtered out the `profileCodeId` null values

andymarke (approved):
lgtm

PR #1042, opened 2025-01-21
SDIT-2480 Check audit module name rather than program service code

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/CourseActivityRepository.kt
```diff
@@ -19,7 +19,7 @@ interface CourseActivityRepository : JpaRepository<CourseActivity, Long> {
     where ca.prison.id = :prisonId
     and ca.active = true
     and (ca.scheduleEndDate is null or ca.scheduleEndDate > current_date) 
-    and ca.program.programCode not in :excludeProgramCodes
+    and ca.auditModuleName != 'DPS_SYNCHRONISATION'
```
mikehalmamoj, line 22: When querying over NOMIS course activities that are eligible for migration we used to check that the program code isn't a DPS program code. However in real life we've found that prisons are using the DPS program codes for course activities created in NOMIS so this caused problems. To fix  and make simpler we'll just check if DPS created the course activity, and if so then this activity is not eligible for migration.

andymarke (approved):
lgtm

PR #1044, opened 2025-01-21
SDIT-2481 Case notes reconciliation username check can fail

steverendell:
Return all usernames for staff id


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/casenotes/CaseNotesService.kt
```diff
@@ -176,7 +177,7 @@ class CaseNotesService(
    * For reconciliation or migration
    */
   fun getCaseNotes(offenderNo: String): PrisonerCaseNotesResponse {
-    offenderRepository.findByNomsId(offenderNo).ifEmpty {
+    if (!offenderRepository.existsByNomsId(offenderNo)) {
```
steverendell, line 180: More efficient

petergphillips (approved):
LGTM

PR #1049, opened 2025-01-22
SDIT-2396: ✨ Map create date on offender too

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/coreperson/CorePersonService.kt
```diff
@@ -44,6 +44,7 @@ class CorePersonService(
           ethnicity = a.ethnicity?.toCodeDescription(),
           sex = a.gender.toCodeDescription(),
           nameType = a.nameType?.toCodeDescription(),
+          createDate = a.createDate,
```
petergphillips, line 47: `createDate` already existed on the JPA offender, so just need to return it

karenmillermoj: Do we need it - or just returning because it exists?

karenmillermoj (approved):
lgtm

PR #1051, opened 2025-01-24
SDIT-2481 Case notes reconciliation username check can fail - duplicate usernames

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/casenotes/CaseNotesService.kt
```diff
@@ -118,7 +118,7 @@ class CaseNotesService(
     // see https://mojdt.slack.com/archives/C06G85DCF8T/p1726158063333349?thread_ts=1726156937.043299&cid=C06G85DCF8T
     authorFirstName = author.firstName,
     authorLastName = author.lastName,
-    authorUsernames = author.accounts.map { it.username },
+    authorUsernames = author.accounts.map { it.username }.distinct(),
     prisonId = agencyLocation?.id ?: offenderBooking.location?.id,
```
steverendell, line 122: The introduction of an entity graph has caused this array to be populated with the username once for every casenote that refers to it via staff id. e.g. in the test below the staff id has 2 usernames and there are 5 case notes, thus the array contains 5 * "JANE.NARK" plus 5 * "JANE.NARK_ADM". For large datasets (prisoners with 1000s of casenotes) this meant the endpoint response body sometimes is > 10M

petergphillips (approved):
LGTM

PR #1060, opened 2025-01-28
SDIT-2519 allow new court event charges during appearance updates. Re…

lbennett-moj:
…move updates during appearance updates


andymarke (approved):
Lgtm

PR #1070, opened 2025-01-30
SDIT-2423 Migrate activities with allocations in the last 6 months

mikehalmamoj:

PR #1073, opened 2025-01-30
Revert "SDIT-2423 Migrate activities with allocations in the last 6 months (#1070)"

mikehalmamoj:
Reverting because I did some before and after tests one of the endpoints in preprod for a handful of prisons, and in some cases there were a LOT of additional activities included in the migration. I need to investigate these first before going live.

Thinking about it I should have just run the endpoint against preprod locally... I'll do that before putting this change back in.

This reverts commit d8ea7713b29508536c756aac8a4dce52d4b0adc5.


andymarke (approved):
lgtm

PR #1079, opened 2025-02-03
SDIT-2547: ✨ New endpoint for reconciliation

petergphillips:


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/casenotes/CaseNotesResourceIntTest.kt
```diff
@@ -677,7 +853,7 @@ class CaseNotesResourceIntTest : IntegrationTestBase() {
           .jsonPath("caseNotes[0].caseNoteText").isEqualTo("Note 1")
           .jsonPath("caseNotes[1].caseNoteId").isEqualTo(id2)
           .jsonPath("caseNotes[1].bookingId").isEqualTo(latestBookingIdA1234AB)
-          .jsonPath("caseNotes[1].caseNoteText").isEqualTo("Note 2")
+          .jsonPath("caseNotes[1].caseNoteText").isEqualTo("basic text ...[JMORROW_GEN updated the case notes on 2023/03/02 17:11:41] made a change")
```
petergphillips, line 856: for reconciliation endpoint we leave the original NOMIS case note text as is and don't separate into amendments

PR #1084, opened 2025-02-10
SDIT-2553 Add event sub type to appointment comment on create

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/appointments/AppointmentService.kt
```diff
@@ -155,6 +155,13 @@ class AppointmentService(
     val eventSubType = eventSubTypeRepository.findByIdOrNull(EventSubType.pk(dto.eventSubType))
       ?: throw BadDataException("EventSubType with code=${dto.eventSubType} does not exist")
 
+    val subTypeDescription = eventSubType.description.trim()
+    val comment = when {
+      dto.comment == null -> subTypeDescription
+      dto.comment.startsWith(subTypeDescription) -> dto.comment
+      else -> "$subTypeDescription - ${dto.comment}".take(4000)
+    }
```
mikehalmamoj, line 163: The nomis-update is to stop adding event sub type description to the comment because DPS are planning to remove that field from their API.

Therefore we will prepend the event sub type description to the comment here - but for now we'll check it's not already there (so we don't have to coordinate releases with DPS).

lbennett-moj (approved):

PR #1087, opened 2025-02-11
SDIT-2553 revert appointment comment changes

mikehalmamoj:
I thought DPS were going to stop sending the customName field to us in all cases - in which case we'd need to add the internal schedule reason to the comment on our side. Turns out they are only going to stop sending custom name if it's the same as the internal schedule reason - so we have nothing to do on our side


andymarke (approved):
lgtm

PR #1100, opened 2025-02-24
SDIT-2575 Make profile details resource global

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/profiledetails/ProfileDetailsResource.kt
```diff
@@ -1,4 +1,4 @@
-package uk.gov.justice.digital.hmpps.nomisprisonerapi.prisonperson.profiledetails
+package uk.gov.justice.digital.hmpps.nomisprisonerapi.profiledetails
```
mikehalmamoj, line 1: The main change is to move profile details out of the prisonperson package and up to the root package. This is because profile details will span multiple data domains so shouldn't be project related.

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/profiledetails/ProfileDetailsResponse.kt
```diff
@@ -18,10 +18,6 @@ data class PrisonerProfileDetailsResponse(
 data class BookingProfileDetailsResponse(
   @Schema(description = "The booking's unique identifier", example = "1234567")
   val bookingId: Long,
-  @Schema(description = "The start date of the booking", example = "2020-07-17T12:34:56")
-  val startDateTime: LocalDateTime,
-  @Schema(description = "The end date of the booking, or null if the booking is still active", example = "2021-07-16T12:34:56")
-  val endDateTime: LocalDateTime?,
```
mikehalmamoj, line 24: Removed the booking start and end dates. Clients are only interested in whether these profile details are for the latest booking or a historical booking. If requirements change we can add back in later.

petergphillips (approved):
LGTM

PR #1103, opened 2025-02-24
SDIT-2954 event id provided with sentence to identify court order date

lbennett-moj:


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingResourceIntTest.kt
```diff
@@ -3807,64 +3826,20 @@ class SentencingResourceIntTest : IntegrationTestBase() {
           .jsonPath("offenderCharges[1].offenceDate").isEqualTo(aLaterDateString)
       }
 
-      @Test
-      fun `can create a sentence without a court order`() {
-        // update appearance to have one inactive charge (removing the court order)
```
mikehalmamoj, line 3812: I was kind of expecting new tests for the new exceptions that are thrown - or maybe something like this test to be converted to a negative test?

mikehalmamoj (approved):

PR #1108, opened 2025-03-03
SDIT-2518 Zero days adjustments - keep status

andymarke:
Original fix was to toggle the status of adjustments that are 0 days to inactive so the DPS CRDS ignores them. This caused a secondary issue for some adjustments; instead DPS CRDS will be changed to ignore the TO DATE for zero day adjustments. The TO DATE will continue not to be written for zero day adjustments since that makes more sense than have a TO DATE before the FROM DATE that was the previous functionality.


petergphillips (approved):
LGTM

PR #1128, opened 2025-03-14
SDIT-2494 Add review sequence to CSIP review create/update

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/csip/CSIPService.kt
```diff
@@ -321,7 +322,7 @@ class CSIPService(
             closeDate = request.closeDate,
             recordedDate = request.recordedDate,
             recordedUser = request.recordedBy,
-            reviewSequence = reviewRepository.getNextSequence(this.id),
+            reviewSequence = request.reviewSequence,
```
karenmillermoj, line 325: The sequence is not part of the primary key - it is just to order the list of reviews

petergphillips (approved):
LGTM

PR #1141, opened 2025-03-24
SDIT-2646 There is no need to update future attendances

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/ActivityService.kt
```diff
@@ -194,14 +194,7 @@ class ActivityService(
...
         offenderProgramProfiles
           .filterNot { it.isEnded() }
-          .forEach {
-            it.program = requestedProgramService
-            val attendanceDetails = it.offenderCourseAttendances.map { it.eventId to it.eventDate }
-            log.info("Updating to program ${requestedProgramService.programCode} (${requestedProgramService.programId}) and considering the following course attendances: $attendanceDetails")
-            it.offenderCourseAttendances.filter { it.eventDate > LocalDate.now() }.forEach { attendance ->
-              attendance.program = program
-            }
-          }
```
mikehalmamoj, line 204: Upon further investigation I realised there is no problem with attendances as they are created on the morning of the attendance so always get the latest program code.

So the only actual problem was with allocations with a future end date which weren't having their program codes updated - and this was fixed in a previous PR.

petergphillips (approved):
LGTM

PR #1157, opened 2025-04-25
SDIT-2726: 🐛 Cope with all prisons too

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/core/ServiceAgencySwitchesResource.kt
```diff
@@ -136,7 +138,7 @@ class ServiceAgencySwitchesResource(private val service: ServiceAgencySwitchesSe
       ),
       ApiResponse(
         responseCode = "403",
-        description = "Forbidden, requires role NOMIS_PRISONER_API__SYNCHRONISATION__RW",
+        description = "Forbidden, requires role NOMIS_ACTIVITIES",
```
petergphillips, line 141: this was wrong from previous PR

build.gradle.kts
```diff
@@ -41,7 +41,7 @@ dependencies {
 kotlin {
   jvmToolchain(21)
   compilerOptions {
-    freeCompilerArgs.add("-Xwhen-guards")
+    freeCompilerArgs.addAll("-Xjvm-default=all", "-Xwhen-guards")
```
petergphillips, line 44: need the jvm default so that the interfaces get generated with a default implementation and can bodies can then be specified

karenmillermoj (approved):
lgtm

PR #1162, opened 2025-05-01
SDIT-2723 Endpoint to return contact

andymarke:
Required so when adding contacts we don't need to call getPerson that doesn't peform well when the person has 1000s of contacts


karenmillermoj (approved):
lgtm

PR #1184, opened 2025-05-16
SDIT-2772 Link case transaction mapping

andymarke:
The link case transaction is returned linked the court event charge that was created


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/NomisData.kt
```diff
@@ -412,6 +418,27 @@ class NomisData(
           builder.apply(dsl)
         }
     }
+
```
andymarke, line 421: added this under a global since it requires multiple cases and yet isn't directly linked to a booking or offender

PR #1188, opened 2025-05-20
SDIT-2764 Changed recall to accept a list of recall sentence types

andymarke:
When recalling not all sentences get converted to the same type, so this is a breaking change.

Also changed to POST since event though the implementation is a update it is actually more like a created. Also we need to PUT for update the recall.

Kept deprecated version until I update the sync service. Converted to test to only test new endpoint

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingService.kt
@@ -1072,6 +1072,65 @@ class SentencingService(
       null,
     )
   }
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingService.kt
```diff
@@ -1072,6 +1072,65 @@ class SentencingService(
       null,
     )
   }
+  fun convertToRecallSentences(offenderNo: String, request: ConvertToRecallRequest) {
```
andymarke, line 1075: More or less as the same as above except each sentence now has its own sentence type.

Wil remove above function after the endpoint is deleted

petergphillips (approved):
LGTM

PR #1195, opened 2025-05-27
SDIT-2764 Add PUT recall and PUT /recall/restore-original

andymarke:
It would be nice to have POST, PUT, DELETE but since in all 3 DPS recall operations in each case we are actually updating existing sentences in a batch; so we have POST and two versions of PUT operations with a slightly different bodies.

 Also, refactor to reduce duplicate code

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingResource.kt
@@ -750,6 +750,168 @@ class SentencingResource(private val sentencingService: SentencingService) {
   )
 
   @PreAuthorize("hasRole('ROLE_NOMIS_SENTENCING')")
+  @PutMapping("/prisoners/{offenderNo}/sentences/recall")
+  @Operation(
+    summary = "Updates Recalls Sentences",
+    description = "Required role NOMIS_SENTENCING Recalls sentences for the offender",
+    requestBody = io.swagger.v3.oas.annotations.parameters.RequestBody(
+      content = [
+        Content(
+          mediaType = "application/json",
+          schema = Schema(implementation = ConvertToRecallRequest::class),
+        ),
+      ],
+    ),
+    responses = [
+      ApiResponse(
+        responseCode = "200",
+        description = "Recall Sentences updated",
+      ),
+      ApiResponse(
+        responseCode = "400",
+        description = "Supplied data is invalid, for instance missing required fields or invalid values. See schema for details",
+        content = [
+          Content(
+            mediaType = "application/json",
+            schema = Schema(implementation = ErrorResponse::class),
+          ),
+        ],
+      ),
+      ApiResponse(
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingResource.kt
```diff
@@ -749,6 +749,168 @@ class SentencingResource(private val sentencingService: SentencingService) {
...
+    offenderNo: String,
+    @RequestBody @Valid
+    request: ConvertToRecallRequest,
+  ) = sentencingService.updateRecallSentences(
+    offenderNo = offenderNo,
+    request = request,
+  )
+
+  @PreAuthorize("hasRole('ROLE_NOMIS_SENTENCING')")
+  @PutMapping("/prisoners/{offenderNo}/sentences/recall/restore-original")
```
andymarke, line 834: NB this can't be a DELETE since it has a body to replace the recall sentences with the original sentence

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingService.kt
```diff
@@ -1019,59 +1022,112 @@ class SentencingService(
     // it would make sense to not make any assumptions
     val bookingIds = request.sentences.map { it.sentenceId.offenderBookingId }.toSet()
 
-    // Update sentences
```
andymarke, line 1022: extracted to function for reuse

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingService.kt
```diff
@@ -1019,59 +1022,112 @@ class SentencingService(
...
+              returnToCustodyDate = this@createOrUpdateBooking.returnToCustodyDate
               staff = enteredByStaff
-              recallLength = request.returnToCustody.recallLength.toLong()
+              recallLength = this@createOrUpdateBooking.recallLength.toLong()
             }
           }
         }
       }
+    } else {
+      bookingIds.forEach { bookingId ->
```
andymarke, line 1111: Interesting I had to "delete" both ends of this one-to-one which makes think perhaps I have the mapping wrong 🤔

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingResourceIntTest.kt
```diff
@@ -5508,7 +5508,7 @@ class SentencingResourceIntTest : IntegrationTestBase() {
   )
 
   @Nested
-  @DisplayName("PUT /prisoners/{offenderNo}/sentences/recall")
+  @DisplayName("POST /prisoners/{offenderNo}/sentences/recall")
```
andymarke, line 5511: corrected typo

PR #1200, opened 2025-06-05
SDIT-2749 Return activities without schedule rules

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/api/FindActiveActivityIdsResponse.kt
```diff
@@ -8,4 +8,6 @@ import io.swagger.v3.oas.annotations.media.Schema
 data class FindActiveActivityIdsResponse(
   @Schema(description = "Activity id", example = "1")
   val courseActivityId: Long,
+  @Schema(description = "Does the activity have schedule rules?", example = "true")
+  val hasScheduleRules: Boolean,
 )
```
mikehalmamoj, line 13: The plan is to include activities in the migration that will be ignored due to missing schedule rules, but we'll still save the mappings for them so we can end the activities.

PR #1209, opened 2025-06-11
SDIT-2802: 🐛 Better mappings of oracle date for incidents

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incidents/IncidentService.kt
```diff
@@ -75,10 +75,10 @@ class IncidentService(
 //      this.staffParties = mutableListOf<IncidentStaffParty>()
 //      this.requirements = mutableListOf<IncidentRequirement>()
         this.reportingStaff = reportedStaff.staff
-        this.reportedDate = request.reportedDateTime.toLocalDate()
-        this.reportedTime = request.reportedDateTime.toLocalTime()
-        this.incidentDate = request.incidentDateTime.toLocalDate()
-        this.incidentTime = request.incidentDateTime.toLocalTime()
+        this.reportedDate = request.reportedDateTime
```
petergphillips, line 78: just leave them both as is with date and time portions

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incidents/IncidentService.kt
```diff
@@ -205,7 +205,7 @@ private fun IncidentRequirement.toRequirement() = Requirement(
   staff = recordingStaff.toStaff(),
   sequence = id.requirementSequence,
   comment = comment,
-  date = recordedDate,
+  date = recordedDate.toLocalDate(),
```
petergphillips, line 208: ideally we'd pass the date time back, but that would change the contract.  So for now, just leave as `LocalDate`

PR #1210, opened 2025-06-11
SDIT-2802: 🐛 Add in new recordedDate field

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/IncidentRequirement.kt
```diff
@@ -39,25 +39,12 @@ class IncidentRequirement(
...
-
-  @Column(name = "MODIFY_DATETIME", insertable = false, updatable = false)
-  @Generated
-  var lastModifiedDateTime: LocalDateTime? = null,
-
-  // ---- NOT MAPPED Columns ---- //
-  // All AUDIT data
-
+  @Column(name = "RECORD_DATE")
+  var recordedDate: LocalDateTime,
```
petergphillips, line 46: note that we don't have a `@Generated` here as we want the value from DPS to go into our database

andymarke (approved):
lgtm

PR #1211, opened 2025-06-11
SDIT-2817 Add court events for updates and deletes

andymarke:
* Just change contract
* TODO implement the update and delete


petergphillips (approved):
LGTM

PR #1215, opened 2025-06-11
SDIT-2802: ✨ Upsert incident requirements

petergphillips:


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/Incident.kt
```diff
@@ -185,7 +185,7 @@ class IncidentBuilder(
         recordingStaff = recordingStaff,
         agencyId = agencyId,
         recordedDate = recordedDate,
-        requirementSequence = incident.requirements.size + 1,
+        requirementSequence = incident.requirements.size,
```
petergphillips, line 188: sequence in database starts at `0`, not `1`

PR #1217, opened 2025-06-12
SDIT-2817 recall revert endpoint

andymarke:
Update and delete are slightly different so new endpoint


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingResource.kt
```diff
@@ -959,6 +959,87 @@ class SentencingResource(private val sentencingService: SentencingService) {
...
+        description = "Booking does not exist",
+        content = [
+          Content(
+            mediaType = "application/json",
+            schema = Schema(implementation = ErrorResponse::class),
+          ),
+        ],
+      ),
+      ApiResponse(
+        responseCode = "404",
```
mikehalmamoj, line 1021: You have 2 of these

PR #1222, opened 2025-06-18
SDIT-2815 Add Nomis Finance endpoints for to-nomis sync - get

steverendell:


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/OffenderTransaction.kt
```diff
@@ -0,0 +1,82 @@
...
+
+  fun build(
+    booking: OffenderBooking,
+    offender: Offender,
+    prisonId: String,
+    transactionType: String,
+  ): OffenderTransaction = OffenderTransaction(
+    offenderBooking = booking,
+    transactionEntrySequence = 1,
+    trustAccount = lookupOrCreateTrustAccount(prisonId, offender),
```
mikehalmamoj, line 61: It feels like there should be a list of trust accounts held against the offender? Though I guess you don't need that if you're not synchronising trust accounts.

PR #1223, opened 2025-06-18
SDIT-2802: ✨ Upsert incident responses

petergphillips:


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/IncidentQuestion.kt
```diff
@@ -57,7 +61,7 @@ class IncidentQuestionBuilder(
     .let { builder ->
       builder.build(
         incidentQuestion = incidentQuestion,
-        answerSequence = incidentQuestion.responses.size + 1,
+        answerSequence = incident.questions.sumOf { it.responses.count() },
```
petergphillips, line 64: note that incident response sequence is across *all* of the questions, not just the current question

PR #1229, opened 2025-06-23
SDIT-2484 removing nomis bug work around after data fix

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingService.kt
```diff
@@ -1550,7 +1550,7 @@ fun OffenderSentence.toSentenceResponse(): SentenceResponse = SentenceResponse(
   createdByUsername = this.createUsername,
   sentenceTerms = this.offenderSentenceTerms.map { it.toSentenceTermResponse() },
   offenderCharges = this.offenderSentenceCharges.map { it.offenderCharge.toOffenderCharge() },
-  missingCourtOffenderChargeIds = getMissingCourtEventChargeIds(),
+  missingCourtOffenderChargeIds = emptyList(),
```
lbennett-moj, line 1553: I'll removing the client code before taking this out

PR #1231, opened 2025-06-24
SDIT-2888 offender restrictions mapping and GET endpoint

andymarke:


src/main/resources/db/migration/V1_262__OFFENDER_RESTRICTIONS.sql
```diff
@@ -1,34 +1,71 @@
+create table OFFENDER_RESTRICTIONS
```
andymarke, line 1: exported latest version for Oracle

PR #1238, opened 2025-06-26
SDIT-2872 Refactor OffenderIndividualSchedule to cater for more than just appointments

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderIndividualSchedule.kt
```diff
@@ -32,7 +34,9 @@ enum class EventClass {
 
 @Entity
 @Table(name = "OFFENDER_IND_SCHEDULES")
-class OffenderIndividualSchedule(
+@DiscriminatorColumn(name = "EVENT_CLASS")
+@Inheritance()
+abstract class OffenderIndividualSchedule(
```
mikehalmamoj, line 39: The main change is to turn this into the parent class and make `OffenderAppointment` a subclass. We can then expand to other subclasses for external movements.

PR #1240, opened 2025-06-27
SDIT-2909 - Court case returns both target and list of source ids

andymarke:
Source case can point at a single target but many sources can point at the same target.

This change models that returns back to the sync service


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/CourtCase.kt
```diff
@@ -80,9 +79,12 @@ class CourtCase(
   @Column(name = "CASE_INFO_NUMBER")
   val primaryCaseInfoNumber: String? = null,
 
-  @OneToOne(fetch = FetchType.LAZY)
+  @ManyToOne(fetch = FetchType.LAZY)
   @JoinColumn(name = "COMBINED_CASE_ID")
-  val combinedCase: CourtCase? = null,
+  val targetCombinedCase: CourtCase? = null,
```
andymarke, line 84: renamed for clarity

steverendell (approved):

PR #1242, opened 2025-06-30
SDIT-2906 Don't update allocation status on end

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/activities/AllocationService.kt
```diff
@@ -330,8 +330,8 @@ class AllocationService(
   fun endAllocation(courseAllocation: OffenderProgramProfile, date: LocalDate, endAllocationComment: String? = null) {
     courseAllocation.endDate = date
     courseAllocation.endReason = findEndReasonOrThrow("OTH")
-    courseAllocation.programStatus = findProgramStatus("END")
     endAllocationComment?.run { courseAllocation.endComment = endAllocationComment }
+    // Note that we don't set the programStatus to END, it's up to NOMIS to manage that status on the correct day
```
mikehalmamoj, line 334: The problem is that if we migrate with a future start date then setting the allocation status to END prevents any attendances from being created in NOMIS. Therefore we decided the solution is just to set the end date and let NOMIS update the status (DPS is only interested in the end date).

PR #1244, opened 2025-06-30
SDIT-2908 Add offenderNo to restriction response

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/contactperson/ContactPersonResource.kt
```diff
@@ -2520,7 +2522,7 @@ data class PrisonerRestriction(
   @Schema(description = "Staff member who created the restriction")
   val enteredStaff: ContactRestrictionEnteredStaff,
   @Schema(description = "Staff member who authorised the restriction")
-  val authorisedStaff: ContactRestrictionEnteredStaff?,
+  val authorisedStaff: ContactRestrictionEnteredStaff,
   @Schema(description = "Audit data associated with the records")
```
andymarke, line 2526: never null on database

PR #1257, opened 2025-07-03
SDIT-2872 Add address to temporary absence applications

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderMovementApplication.kt
```diff
@@ -118,12 +118,13 @@ class OffenderMovementApplication(
...
-//
-//  @Column(name = "TO_ADDRESS_ID")
-//  val toAddressId: Long? = null,
+
+  @Column(name = "TO_ADDRESS_OWNER_CLASS")
+  val toAddressOwnerClass: String? = null,
+
+  @ManyToOne(fetch = FetchType.LAZY)
+  @JoinColumn(name = "TO_ADDRESS_ID")
+  val toAddress: Address? = null,
```
mikehalmamoj, line 127: The main change is to support addresses. We can't map the address owner class from the Address because we expect owner class of AGY, CORP or OFF but the OFFENDER_MOVEMENT_APPS table has loads where the address owner class is actually the AGY_LOC_ID. I'll deal with them next.

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/NomisData.kt
```diff
@@ -441,6 +445,58 @@ class NomisData(
           builder.apply(dsl)
         }
     }
+
+  // TODO - I chickened out of mapping AGENCY_LOCATIONS because AGY_LOC_ID is everywhere. Left it for the person who migrates agencies.
+  override fun agencyAddress(
```
mikehalmamoj, line 450: This was a bit naughty - leaving address agency as a top level entity in the test data builder DSL, when it should be a child of agency. However, that looks like a big job unrelated to my needs.

PR #1259, opened 2025-07-04
SDIT-2872 Add temporary absence application date and time

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderMovementApplication.kt
```diff
@@ -57,7 +57,7 @@ class OffenderMovementApplication(
   val eventSubType: MovementReason,
 
   @Column(name = "APPLICATION_DATE")
-  val applicationDate: LocalDate,
+  val applicationDate: LocalDateTime,
 
   @Column(name = "APPLICATION_TIME")
   val applicationTime: LocalDateTime,
```
mikehalmamoj, line 63: Application date and time are both DATE columns in NOMIS. Recently this is only the date part (in both columns) but historically this could include the time (in both columns). So we need to cater for times. In both columns.

PR #1286, opened 2025-07-18
SDIT-2918 Don't include allocations in migration that end before DPS starts

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderProgramProfileRepository.kt
```diff
@@ -46,7 +46,7 @@ interface OffenderProgramProfileRepository : JpaRepository<OffenderProgramProfil
        join CourseScheduleRule csr on ca = csr.courseActivity
        where opp.prison.id = :prisonId 
        and opp.programStatus.code = 'ALLOC'
-       and (opp.endDate is null or opp.endDate > current_date)
+       and (opp.endDate is null or opp.endDate >= :activeOnDate)
```
mikehalmamoj, line 49: We used to always migrate the day before DPS starts - so were looking for allocations that are still active tomorrow. However we now migrate 2 or 3 days before DPS starts. This means we're looking for allocations that are active on the start date - if an allocation ends before DPS starts then if we try to migrate DPS rejects the allocation anyway.

PR #1288, opened 2025-07-23
SDIT-2872 Handle TAP schedules created by merges

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/MovementsService.kt
```diff
@@ -46,6 +48,27 @@ class MovementsService(
...
+
+    // put the scheduled returns onto the correct application
+    scheduledReturnsWithWrongParent.forEach { mergedReturn ->
+      this.flatMap { booking -> booking.temporaryAbsenceApplications }
+        .find { it.movementApplicationId == mergedReturn.temporaryAbsenceApplication?.movementApplicationId }
+        ?.scheduledTemporaryAbsence
+        ?.scheduledTemporaryAbsenceReturns += mergedReturn
+      mergedReturn.temporaryAbsenceApplication = null
+    }
+  }
```
mikehalmamoj, line 70: Before mapping the schedules to the DTO, we move all of the corrupted scheduled returns onto the correct TAP applications, reinstating the one to one mapping.

PR #1294, opened 2025-07-29
SDIT-2872 Provide a basic NOMIS audit entity that doesn't cause n+1 queries on staff

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/NomisAuditableEntityWithStaff.kt
```diff
@@ -9,9 +9,13 @@ import org.hibernate.annotations.Generated
...
 import java.time.LocalDateTime
 
+/*
+ * This includes a join on the staff user accounts table that causes n+1 queries.
+ * Consider using NomisAuditableEntityBasic instead.
+ */
 @MappedSuperclass
 @EntityOpen
-class NomisAuditableEntity {
+class NomisAuditableEntityWithStaff {
```
mikehalmamoj, line 18: Renamed to make it clear that this is including a join on the staff table.

PR #1297, opened 2025-07-29
SDIT-2872 prevent n+1 queries

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/MovementsService.kt
```diff
@@ -30,39 +29,43 @@ class MovementsService(
       throw NotFoundException("Offender with nomsId=$offenderNo not found")
     }
 
-    val bookings = offenderBookingRepository.findAllByOffenderNomsId(offenderNo)
-
-    bookings.fixMergedSchedules()
+    val movementApplications = offenderMovementApplicationRepository.findAllByOffenderBooking_Offender_NomsId(offenderNo)
+      .also { it.fixMergedSchedules() }
+    val unscheduledTemporaryAbsences = temporaryAbsenceRepository.findAllByOffenderBooking_Offender_NomsIdAndScheduledTemporaryAbsenceIsNull(offenderNo)
+    val unscheduledTemporaryAbsenceReturns = temporaryAbsenceReturnRepository.findAllByOffenderBooking_Offender_NomsIdAndScheduledTemporaryAbsenceIsNull(offenderNo)
```
mikehalmamoj, line 35: The main change is to drive the retrieval of applications and unscheduled movements from the offenderNo rather than by getting all bookings as this means there is a single query for each - previously we had n+1 queries on each.

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderMovementApplication.kt
```diff
@@ -10,18 +10,70 @@ import jakarta.persistence.GeneratedValue
...
+  attributeNodes = [
+    NamedAttributeNode(value = "offenderBooking"),
+    NamedAttributeNode(value = "scheduledTemporaryAbsence", subgraph = "scheduled-temporary-absence"),
+    NamedAttributeNode(value = "eventSubType"),
+    NamedAttributeNode(value = "applicationStatus"),
+    NamedAttributeNode(value = "escort"),
+    NamedAttributeNode(value = "transportType"),
+    NamedAttributeNode(value = "applicationType"),
+    NamedAttributeNode(value = "temporaryAbsenceType"),
+    NamedAttributeNode(value = "temporaryAbsenceSubType"),
```
mikehalmamoj, line 41: This means we have to tell hibernate to join on the reference codes too.

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/MovementsResourceIntTest.kt
```diff
@@ -539,34 +539,21 @@ class MovementsResourceIntTest(
...
-
-      // the type and subtype are null after copying from the merged booking
-      jdbcTemplate.update(
-        """
-        update OFFENDER_MOVEMENT_APPS set 
-          TAP_ABS_TYPE=null,
-          TAP_ABS_SUBTYPE=null
-        where OFFENDER_MOVEMENT_APP_ID = ${application.movementApplicationId}
-        """.trimIndent(),
-      )
```
mikehalmamoj, line 554: Dropped this merge related corruption because it was causing problems when saving the transaction with jdbc, couldn't get an update query to work with entityManager due to the JoinFormula, and it doesn't really matter anyway as it's not the interesting part of the corrupted data.

andymarke (approved):
LGTM

PR #1304, opened 2025-08-04
SDIT-2982 Endpoint to clone court cases to latest booking

andymarke:
* Now clones core case data held in DPS

TODO - clone the case children


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingResourceIntTest.kt
```diff
@@ -1959,22 +1991,251 @@ class SentencingResourceIntTest : IntegrationTestBase() {
...
+            assertThat(primaryCaseInfoNumber).isEqualTo("X0001")
+            assertThat(caseSequence).isGreaterThan(1)
+            assertThat(beginDate).isEqualTo(LocalDate.parse("2021-01-02"))
+            assertThat(court.id).isEqualTo("LEICYC")
+            assertThat(legalCaseType.code).isEqualTo("Y")
+            assertThat(caseStatus.code).isEqualTo("I")
+            assertThat(targetCombinedCase).isNull()
+            assertThat(sourceCombinedCases).isEmpty()
+
+            // TODO use default values - no copy for now - but double check this is ok given they are not held in DPS
```
andymarke, line 2215: Just not sure what to do here; data that doesn't exist in DPS but does in NOMIS - not sure if we should copy or not. I'll speak to @lbennett-moj

PR #1320, opened 2025-08-11

andymarke:
Expose bug for linked cases where we cannot clone due to sourceCase no longer pointing at offender charge that is still on a court event

Will then fix bug - but setting up linked cases data is a large job in itself

* Linked cases links
* Adjustments

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/LinkCaseTxn.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/LinkCaseTxn.kt
@@ -1,13 +1,18 @@
 package uk.gov.justice.digital.hmpps.nomisprisonerapi.helper.builders
 
+import org.springframework.data.repository.findByIdOrNull
 import org.springframework.jdbc.core.JdbcTemplate
 import org.springframework.stereotype.Component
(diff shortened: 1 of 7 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/LinkCaseTxn.kt
```diff
@@ -29,17 +34,73 @@ class LinkCaseTxnBuilderFactory(
...
   private val repository: LinkCaseTxnBuilderRepository,
 ) : LinkCaseTxnDsl {
   private lateinit var linkCaseTxn: LinkCaseTxn
+  fun linkCases(
+    sourceCase: CourtCase,
+    targetCase: CourtCase,
+    courtEvent: CourtEvent,
+    whenCreated: LocalDateTime,
+  ): List<LinkCaseTxn> {
+    val chargesToCopy = sourceCase.offenderCharges.filterNot { it.resultCode1?.dispositionCode == "F" }
```
andymarke, line 65: Simulate what NOMIS would do when linking cases
1. Move charges from source to target that are non-final
2. Added charges to target court event
3. Create linkCaseTxn (which just could not get JPA to do it - due to the OffenderCharge changing parent maybe??)

PR #1322, opened 2025-08-11

andymarke:
Linked cases links aka combined cases.

Complicated by the courtCase -> courtEvent -> courtEventCharge -> offenderCharge does not play well when the offenderCharge is pointed at by courtEventCharge in both the source and target linked cases. Then further complicated by the LinkCaseTxn entities point at all of the above making transient entity exceptions hard to avoid. Then further complicated that none of this can be persisted until we have persisted all the about objects for every case being cloned.

So fallen back to using SQL for inserting LinkCaseTxn

* Case status
* Adjustments

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingService.kt
@@ -25,6 +25,7 @@ import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.DirectionType
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.EventStatus
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.LegalCaseType
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.LinkCaseTxn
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingService.kt
```diff
@@ -1444,29 +1446,6 @@ class SentencingService(
                 }
               }
             }
-            clonedCourtEvent.courtEventCharges += courtEvent.courtEventCharges.map { courtEventCharge ->
-              CourtEventCharge(
```
andymarke, line 1448: moved down to after courtEvents have been flushed

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingResourceIntTest.kt
```diff
@@ -2355,6 +2355,7 @@ class SentencingResourceIntTest : IntegrationTestBase() {
                 statusUpdateStaff = staff,
                 caseInfoNumber = "SOURCE",
                 caseSequence = 1,
+                caseStatus = "I",
               ) {
```
andymarke, line 2359: Will show how these changes when closed in next PR

PR #1326, opened 2025-08-12
SDIT-2982 Add nullable clonedCourtCases to the court appearance result.

andymarke:
TODO: scenario when the court appearance is added to a case on an old booking

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisoners/BookingResourceIntTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisoners/BookingResourceIntTest.kt
@@ -174,4 +174,4 @@ class BookingResourceIntTest : IntegrationTestBase() {
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisoners/BookingResourceIntTest.kt
```diff
@@ -174,4 +174,4 @@ class BookingResourceIntTest : IntegrationTestBase() {
   }
 }
 
-inline fun <reified B> WebTestClient.ResponseSpec.expectBodyResponse(): B = this.expectStatus().isOk.expectBody(B::class.java).returnResult().responseBody!!
+inline fun <reified B> WebTestClient.ResponseSpec.expectBodyResponse(): B = this.expectStatus().is2xxSuccessful.expectBody(B::class.java).returnResult().responseBody!!
```
andymarke, line 177: nice to be a lenient here. Test consumers expecting an 200 response won't case so long is it is a 2xx result - but you cam still independently test for say a 201 if you care (see below)

PR #1329, opened 2025-08-13
SDIT-2962 Transaction reconciliation

steverendell:
Endpoints for reconciliation process for finance

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/finance/TransactionsResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/finance/TransactionsResource.kt
@@ -11,9 +11,11 @@ import org.springframework.validation.annotation.Validated
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.ResponseStatus
 import org.springframework.web.bind.annotation.RestController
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/finance/TransactionsResource.kt
```diff
@@ -81,6 +83,99 @@ class TransactionsResource(
...
+      generalLedgerEntrySequence,
+      pageSize,
+    )
+  // TODO not sure if needed at some point:
+//    if (all.size < pageSize) {
+//      return all
+//    }
+//    val maxId = all.last().transactionId // all.maxBy { it.transactionId }.transactionId
+//    return all
+//      .filter { it.transactionId < maxId }
```
steverendell, line 178: This is for transactions not linked to any prisoner. The idea here is translate a start date (eg 1 month ago) to a transaction id, then iterate over transactions by id, page by page, using a 'last' transaction id/seq/glseq combo

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/finance/FinanceResourceIntTest.kt
```diff
@@ -205,4 +206,179 @@ class FinanceResourceIntTest : IntegrationTestBase() {
...
+          .expectStatus().isUnauthorized
+      }
+    }
+
+    @Nested
+    inner class HappyPath {
+      @BeforeEach
+      fun setUp() {
+        nomisDataBuilder.build {
+          generalLedgerTransaction(10, 1, 1)
```
andymarke, line 297: it is worth adding parameter names here - especially as they are all the same type

PR #1343, opened 2025-08-19
SDIT-2880 Standardise role for /agency-switches endpoints

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/core/ServiceAgencySwitchesResource.kt
```diff
@@ -161,12 +159,11 @@ class ServiceAgencySwitchesResource(private val service: ServiceAgencySwitchesSe
     }
   }
 
-  @PreAuthorize("hasRole('ROLE_NOMIS_ACTIVITIES')")
   @PostMapping("{serviceCode}/agency/{agencyId}")
```
karenmillermoj, line 162: I have updated the dashboard-client that uses this endpoint, with NOMIS_PRISONER_API__SYNCHRONISATION__RW role in all 3 envs.

PR #1344, opened 2025-08-20
SDIT-2860 GET movement application

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/MovementsResource.kt
```diff
@@ -55,4 +55,42 @@ class MovementsResource(
   fun getTemporaryAbsencesAndMovements(
     @Schema(description = "Offender number (NOMS ID)", example = "A1234BC") @PathVariable offenderNo: String,
   ): OffenderTemporaryAbsencesResponse = movementsService.getTemporaryAbsencesAndMovements(offenderNo)
+
+  @PreAuthorize("hasRole('ROLE_NOMIS_MOVEMENTS')")
+  @GetMapping("/movements/{offenderNo}/temporary-absences/application/{applicationId}")
```
mikehalmamoj, line 60: New endpoint that gets a movement application only - not any of the related schedules or movements themselves. This is because we're syncing fine grained entities, not families of entities.

PR #1361, opened 2025-08-26
SDIT-2982 Clone court cases

andymarke:
Also clone any related sentence adjustments and make them active

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingAdjustmentService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingAdjustmentService.kt
@@ -290,6 +290,32 @@ class SentencingAdjustmentService(
     )
   }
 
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingAdjustmentService.kt
```diff
@@ -290,6 +290,32 @@ class SentencingAdjustmentService(
     )
   }
 
+  @Transactional(propagation = Propagation.MANDATORY)
+  fun copyAllAdjustment(fromSentences: List<OffenderSentence>, toSentences: List<OffenderSentence>): List<SentenceIdAndAdjustments> = fromSentences.zip(toSentences).map { (fromSentence, toSentence) ->
```
mikehalmamoj, line 294: This assumes that fromSentences and toSentences match and hence can be zipped - just wondering if that assume is valid?

mikehalmamoj (approved):

PR #1402, opened 2025-09-18
SDIT-3048 Ensure hashcode/equals set for TrustAccount and improve readability of sql

karenmillermoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/csip/CSIPFactorResourceIntTest.kt
```diff
@@ -34,10 +34,10 @@ class CSIPFactorResourceIntTest : IntegrationTestBase() {
             releaseDate = LocalDate.parse("2028-11-25"),
           ) {
             factor()
-            factor(factor = "GAN")
```
karenmillermoj, line 37: IntelliJ issues - renamed to ensure match correct override field

PR #1419, opened 2025-09-26
SDIT-3066: 🐛 Try out fix for pageable swagger generation

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/finance/PrisonerBalanceResource.kt
```diff
@@ -57,8 +58,9 @@ class PrisonerBalanceResource(
       ),
     ],
   )
-  fun getPrisonerIdentifiers(
+  fun getPrisonerBalanceIdentifiers(
```
petergphillips, line 61: rename method so unique and doesn't then get a number added to the function e.g. `getPrisonerIdentifiers1RequestConfig`

karenmillermoj (approved):
lgtm

PR #1420, opened 2025-09-26
SDIT-3064 Ensure only one order is created per court appearance.

andymarke:
When multiple update charge events are received for a single appearance at the same time there is a race conditioning meaning multiple orders can be created for the same appearance. This is legal in NOMIS but not quite what we want.

Allow reading of the first order when there are multiples.

NOMIS allows multiple orders per event, but it is rare. We ensure we don't create multiples when several update events are received at the same time by using a SQL insert with a `where` clause on event_id

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/CourtOrderRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/CourtOrderRepository.kt
@@ -8,7 +8,7 @@ import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.OffenderBooking
 
 @Repository
 interface CourtOrderRepository : JpaRepository<CourtOrder, Long> {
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/CourtOrderRepository.kt
```diff
@@ -8,7 +8,7 @@ import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.OffenderBooking
 
 @Repository
 interface CourtOrderRepository : JpaRepository<CourtOrder, Long> {
-  fun findByOffenderBookingAndCourtEventAndOrderType(
+  fun findFirstByOffenderBookingAndCourtEventAndOrderTypeOrderByIdAsc(
     offenderBooking: OffenderBooking,
```
andymarke, line 12: there are 115 duplicates in live; just select the first one - it actually doesn't matter which one is selected so long as there is one

PR #1422, opened 2025-10-01
SDIT-3015 Endpoint for standalone case clone endpoint

andymarke:
Will be used for repairing a case that has not been cloned


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/sentencing/SentencingResourceIntTest.kt
```diff
@@ -3364,6 +3364,1381 @@ class SentencingResourceIntTest : IntegrationTestBase() {
     }
   }
 
+  @Nested
```
andymarke, line 3367: This test is a more less copy in spirit of existing test that copies all cases in a booking (which itself is not used and will be deleted in next PR)

mikehalmamoj (approved):

PR #1704, opened 2026-04-09
SDIT-3666 Add cloning source info to audit

andymarke:
When cloning a court case an children add info to AUDIT_ADDITIONAL_INFO indicating this is a copy. This will aid debugging and will also be used to distinguish when to sync services that a new enity is in fact just a clone


steverendell (approved):
lgtm

PR #1705, opened 2026-04-09
SDIT-3666 Add clone flag for court appearance

andymarke:
Indicates that a court appearance was marked as clone created from the court case being copied to latest booking


petergphillips (approved):
LGTM

PR #1706, opened 2026-04-09
SDIT-3666 clone use additional audit

andymarke:
1. Revert original manual setting of AUDIT_ADDITIONAL_INFO that does not work - it has to be done via set_context
2. Keep flag (which will always be force) ready for fix to AUDIT_ADDITIONAL_INFO

TODO - use set_context to set AUDIT_ADDITIONAL_INFO

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
@@ -2,7 +2,6 @@ package uk.gov.justice.digital.hmpps.nomisprisonerapi.courtsentencing
 
 import com.microsoft.applicationinsights.TelemetryClient
 import org.slf4j.LoggerFactory
-import org.springframework.beans.factory.annotation.Value
 import org.springframework.data.domain.Page
 import org.springframework.data.domain.Pageable
 import org.springframework.data.repository.findByIdOrNull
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/CourtEvent.kt
```diff
@@ -133,9 +134,9 @@ class CourtEvent(
     OUTCOME_DATE - not used
     OFFENDER_PROCEEDING_ID - not used
    */
-  @Column(name = "AUDIT_ADDITIONAL_INFO")
-  val auditAdditionalInfo: String? = null,
-
+  @Column(name = "AUDIT_ADDITIONAL_INFO", insertable = false, updatable = false)
```
andymarke, line 137: ready for next PR when this is populated - but this is always a generated field since populated via trigger

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
```diff
@@ -2171,7 +2163,7 @@ private fun CourtEvent.toCourtEvent(): CourtEventResponse = CourtEventResponse(
   modifiedByUsername = this.modifyUserId,
   courtEventCharges = this.courtEventCharges.map { it.toCourtEventCharge() },
   courtOrders = this.courtOrders.map { it.toCourtOrder() },
-  isClone = this.auditAdditionalInfo?.startsWith("DPS Cloned from") == true,
+  isClone = this.auditAdditionalInfo?.startsWith("DPS Cloned") == true,
 )
```
andymarke, line 2167: Ready for next PR when we set text to "DPS Cloned" in AUDIT_ADDITIONAL_INFO

petergphillips (approved):
LGTM

PR #1709, opened 2026-04-10
SDIT-2832 Refactor TAP API

mikehalmamoj:
Created a new API for TAPs. This is a **copy** of the old API with everything renamed to the new naming convention (URLs, model, class name).

The old API is still there for now. The plan is to switch clients to the new API before removing the old API.


petergphillips (approved):
LGTM

PR #1714, opened 2026-04-14
SDIT-3678 Lock incident on update

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incidents/IncidentService.kt
```diff
@@ -76,7 +76,7 @@ class IncidentService(
     val reportedStaff = lookupStaff(request.reportedBy)
 
     return (
-      incidentRepository.findByIdOrNull(incidentId)?.apply {
+      incidentRepository.findByIdOrNullForUpdate(incidentId)?.apply {
```
karenmillermoj, line 79: Just adding the lock on the update initially (rather than on the quick create /save and then update) - to check child table locking first

petergphillips (approved):
LGTM

PR #1717, opened 2026-04-20
SDIT-3678 Identify duplicate incident inserts and throw locked exception

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incidents/IncidentService.kt
```diff
@@ -112,7 +115,16 @@ class IncidentService(
       if (request.history.size > it.incidentHistory.size) {
         insertHistory(it.incidentHistory, request.history.drop(it.incidentHistory.size))
       }
-      incidentRepository.save(it)
+
```
karenmillermoj, line 118: This will allow us to identify the duplicate inserts rather than it throwing a 500 error, as it currently does.  Once live and exceptions coming thru successfully - I can then write a specific retry in the to-nomis service.

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incidents/IncidentService.kt
```diff
@@ -112,7 +115,16 @@ class IncidentService(
...
+      try {
+        incidentRepository.save(it)
+      } catch (e: DataIntegrityViolationException) {
+        if (e.isIncidentDuplicate()) {
+          log.error("Attempted to create incident that already exists")
```
petergphillips: not sure I'd bother with this `log.error` here since the exception handler will log it anyway

petergphillips (dismissed):
LGTM

petergphillips (approved):
LGTM

PR #1721, opened 2026-04-21
SDIT-3678 Allow for schema when checking for duplicate insert exception

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incidents/IncidentService.kt
```diff
@@ -355,7 +355,7 @@ class IncidentService(
 
   fun DataIntegrityViolationException.isIncidentDuplicate(): Boolean {
     val constraintViolation = cause as? ConstraintViolationException
-    return (constraintViolation != null && constraintViolation.constraintName == "INCIDENT_CASES_PK")
+    return (constraintViolation?.constraintName?.contains("INCIDENT_CASES_PK") == true)
```
karenmillermoj, line 358: the constraint had the schema name too - have changed to contains to ensure it will be picked up

mikehalmamoj (approved):

PR #1730, opened 2026-04-29
SDIT-2832 Create new endpoint for court schedule OUT that does nothing

mikehalmamoj:
Designed the API and model - next step is to implement the service.


steverendell (approved):
lgtm

PR #1733, opened 2026-04-30
SDIT-2832 Implement get court schedule OUT service

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/court/schedule/CourtScheduleService.kt
```diff
@@ -1,20 +1,49 @@
...
-    "any",
-    "any",
-    "any",
-  )
+class CourtScheduleService(
+  private val courtEventRepository: CourtEventRepository,
+  private val externalMovementRepository: OffenderExternalMovementRepository,
+  private val offenderRepository: OffenderRepository,
+) {
+  fun getCourtScheduleOut(offenderNo: String, eventId: Long): CourtScheduleOut {
```
mikehalmamoj, line 20: Main change is to implement the service

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/court/schedule/CourtScheduleService.kt
```diff
@@ -1,20 +1,49 @@
...
+      ?: throw NotFoundException("Court event with id=$eventId not found for prisoner with nomsId=$offenderNo")
+  }
+
+  private fun findPrisonAt(time: LocalDateTime, bookingId: Long): AgencyLocation? {
+    val admissions = externalMovementRepository.findAllById_OffenderBooking_BookingId(bookingId)
+      .filter { it.movementReason.id.type == "ADM" }
+    val admission = admissions.filter { it.movementTime < time }.maxByOrNull { it.movementTime }
+      ?: admissions.minByOrNull { it.movementTime }
+    return admission?.toAgency
+  }
```
mikehalmamoj, line 48: They want the prison at the time the the court schedule was created. I didn't want to copy and paste the get timeline code from prison-api because it's known to be slow and we'd have to call it for every schedule. Instead I copied the logic from it to find the most recent admission before the court schedule create time.

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/OffenderExternalMovement.kt
```diff
@@ -94,7 +94,7 @@ class OffenderExternalMovementBuilder(
     OffenderExternalMovement(
       id = OffenderExternalMovementId(
         offenderBooking,
-        offenderBooking.externalMovements.size + 1,
+        offenderBooking.externalMovements.size + 2,
```
mikehalmamoj, line 97: Unlrelated fix - this was overriding the transfer OUT so you only saw a movement IN.

petergphillips (approved):
LGTM

PR #1736, opened 2026-05-01
SDIT-3723 Add test data builders for court movements

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderExternalMovement.kt
```diff
@@ -45,6 +44,8 @@ data class OffenderExternalMovementId(
     case
         when MOVEMENT_TYPE = 'TAP' and DIRECTION_CODE = 'OUT' then 'OffenderTapMovementOut'
         when MOVEMENT_TYPE = 'TAP' and DIRECTION_CODE = 'IN' then 'OffenderTapMovementIn'
+        when MOVEMENT_TYPE = 'CRT' and DIRECTION_CODE = 'OUT' then 'OffenderCourtMovementOut'
+        when MOVEMENT_TYPE = 'CRT' and DIRECTION_CODE = 'IN' then 'OffenderCourtMovementIn'
```
mikehalmamoj, line 48: The main change to production code is to model the court movements that are extended from external movements as they're on the same table (`OFFENDER_EXTERNAL_MOVEMENTS`)

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/court/CourtMovementResourceIntTest.kt
```diff
@@ -1,23 +1,83 @@
...
+                courtMovementIn()
+              }
+              courtCase(reportingStaff = staff) {
+                courtEvent {
+                  courtMovementOut()
+                  courtMovementIn()
+                }
+              }
+            }
+          }
```
mikehalmamoj, line 56: Everything else exists to support the test data model allowing us to model the different types of court events and movements, which we test below.

steverendell (approved):
lgtm

PR #1742, opened 2026-05-05
SDIT-3755 Hack to allow timeslots to be created and updated

andymarke:
Due to a NOMIS constraint, we have to jump through hoops to try to prevent active and inactive time slots sharing the same start and end time. Though we (and NOMIS) are only interested in the time portion, the DB constraint prevents two or more rows sharing the same DateTime for start and end time for a given day in a prison.

Therefore, new rows created deafult to todays date and any updates retain that original date. This is more consistent to how NOMIS sets the date portion.

NB VSIP (social visits) sync is left unchanged since that sync only sets the time portion with the date portion defaulting to the epoch start date.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AgencyVisitTime.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AgencyVisitTime.kt
@@ -18,6 +18,7 @@ import org.springframework.beans.factory.annotation.Value
 import org.springframework.data.domain.Persistable
 import java.io.Serializable
 import java.time.LocalDate
+import java.time.LocalDateTime
 import java.time.LocalTime
 
 @Entity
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/AgencyVisitTimeRepository.kt
```diff
@@ -61,12 +62,35 @@ interface AgencyVisitTimeRepository : JpaRepository<AgencyVisitTime, AgencyVisit
   @Query(
     """
       update AgencyVisitTime avt
```
andymarke, line 64: use HQL so we can set the startDateTime attributes directly which are updatable / insertable false

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AgencyVisitTime.kt
```diff
@@ -32,6 +33,13 @@ data class AgencyVisitTime(
   @Column(name = "END_TIME", nullable = false)
   var endTime: LocalTime,
 
+  // used to be able to set date element of what is only used by the NOMIS as the time
```
andymarke, line 36: make these read only but can still be set via HQL

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/AgencyVisitDay.kt
```diff
@@ -25,6 +26,16 @@ interface AgencyVisitDayDsl {
     expiryDate: LocalDate? = null,
     dsl: AgencyVisitTimeDsl.() -> Unit = {},
   ): AgencyVisitTime
+
+  @AgencyVisitTimeDslMarker
+  fun visitTimeSlot(
```
andymarke, line 31: a version that allows date parts to be set

karenmillermoj (approved):
lgtm

PR #1755, opened 2026-05-11
SDIT-3723 Check direction for court schedules and movements

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/court/schedule/CourtScheduleService.kt
```diff
@@ -24,8 +25,10 @@ class CourtScheduleService(
     if (!offenderRepository.existsByNomsId(offenderNo)) {
       throw NotFoundException("Offender with nomsId=$offenderNo not found")
     }
-    return courtEventRepository.findByIdOrNull(eventId)?.toResponse()
-      ?: throw NotFoundException("Court event with id=$eventId not found for prisoner with nomsId=$offenderNo")
+    return courtEventRepository.findByIdOrNull(eventId)
+      ?.takeIf { it.directionCode?.code == DirectionType.OUT }
+      ?.toResponse()
+      ?: throw NotFoundException("Court event OUT with id=$eventId not found for prisoner with nomsId=$offenderNo")
   }
```
mikehalmamoj, line 32: For court schedules we currently need to check the direction in the API because it's not included on the XTAG event, so we call this endpoint for all schedules even though we don't sync court schedules IN.

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/court/movement/CourtMovementService.kt
```diff
@@ -28,6 +30,7 @@ class CourtMovementService(
     val offenderBooking = findOffenderBookingOrThrow(offenderNo, bookingId)
 
     return courtMovementOutRepository.findByIdOrNull(OffenderExternalMovementId(offenderBooking, sequence))
+      ?.takeIf { it.movementDirection == OUT }
```
mikehalmamoj, line 33: Might as well add the same validation for court movements too even though it'll probably never be used.

lbennett-moj (approved):

PR #1764, opened 2026-05-14
SDIT-3731 court movements performance

mikehalmamoj:
Reverts the previous revert to remove get offender court movements endpoint due to performance issues.

Then rework the JPA to prevent said performance issues. This mainly involved taking out the @OneToOne link between CourtEvent and OffenderCourtMovementOut/In because it appears that the SQL generated caused the performance issues we saw in production. We don't really need CourtEvent to know about related movements anyway, it was mainly done for completeness (plus that approach seemed to work fine for TAPs).

In fact if you look at the [2nd commit](https://github.com/ministryofjustice/hmpps-nomis-prisoner-api/pull/1764/changes/722fd77506751fb6a096cae635406277d6c730c3) you can see the JPA fixes.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderCourtMovementIn.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderCourtMovementIn.kt
@@ -1,10 +1,10 @@
 package uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa
 
+import jakarta.persistence.Column
 import jakarta.persistence.Entity
 import jakarta.persistence.FetchType.LAZY
 import jakarta.persistence.JoinColumn
 import jakarta.persistence.ManyToOne
-import jakarta.persistence.OneToOne
 import org.hibernate.annotations.NotFound
 import org.hibernate.annotations.NotFoundAction
 import java.time.LocalDate
@@ -21,10 +21,9 @@ class OffenderCourtMovementIn(
   active: Boolean = false,
   commentText: String? = null,
 
-  @OneToOne()
-  @JoinColumn(name = "PARENT_EVENT_ID")
+  @Column(name = "PARENT_EVENT_ID")
   @NotFound(action = NotFoundAction.IGNORE)
-  var courtScheduleOut: CourtEvent? = null,
+  var courtScheduleOutId: Long? = null,
 ) : OffenderExternalMovement(
   id = id,
   movementDate = movementDate,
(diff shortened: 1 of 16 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderCourtMovementIn.kt
```diff
@@ -21,10 +21,9 @@ class OffenderCourtMovementIn(
   active: Boolean = false,
   commentText: String? = null,
 
-  @OneToOne()
-  @JoinColumn(name = "PARENT_EVENT_ID")
+  @Column(name = "PARENT_EVENT_ID")
   @NotFound(action = NotFoundAction.IGNORE)
-  var courtScheduleOut: CourtEvent? = null,
+  var courtScheduleOutId: Long? = null,
```
mikehalmamoj, line 26: Also removed the join from court movements back to the court event - again this is not needed and was mainly done for completeness.

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/court/offender/OffenderCourtMovementsService.kt
```diff
@@ -1,10 +1,109 @@
...
+      throw NotFoundException("Offender with nomsId=$offenderNo not found")
+    }
+
+    val allMovementsOut = courtMovementOutRepository.findAllByOffenderBooking_Offender_NomsId(offenderNo)
+    val allMovementsIn = courtMovementInRepository.findAllByOffenderBooking_Offender_NomsId(offenderNo)
+    val allSchedulesOut = courtEventRepository.findAllByOffenderBooking_Offender_NomsIdAndDirectionCode_CodeIs(offenderNo, "OUT")
+
+    // When finding unscheduled movements we also have to cross reference with the schedules we loaded - just in case any are linked to a schedule without a direction (bad old NOMIS data that we are ignoring)
+    val unscheduledMovementsOut = allMovementsOut.filter { it.courtScheduleOutId == null || it.courtScheduleOutId !in(allSchedulesOut.map { it.id }) }
+    val unscheduledMovementsIn = allMovementsIn.filter { it.courtScheduleOutId == null || it.courtScheduleOutId !in(allSchedulesOut.map { it.id }) }
```
mikehalmamoj, line 37: The main reason we don't need a parent/child relationship between court events and court movements is that due to the lack of referential integrity in NOMIS and the bad data this has allowed, we can't trust that relationship anyway. Which means we have to manually calculate which movements are unscheduled or orphaned and which are genuinely linked to a court event regardless of JPA.

petergphillips (approved):
LGTM

PR #1768, opened 2026-05-18
SDIT-3816 Update charge outcome of appearance update

andymarke:
The bug scenario was as follows:
1. Existing appearance and charges with no-final outcome (e.g. Adjourned)
2. DPS updates appearance and charges to Imprisonment
3. All events fired at same time, court appearance updated and the court charges updated events
4. Court order created by charge update handling
5. Court order deleted by appearance update handling since it has a stale version of the charge outcomes

This ensures the outcomes for existing charges is updated at same time as court appearance so court order logic always has lastest outcome codes.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingResourceIntTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingResourceIntTest.kt
@@ -5269,6 +5269,7 @@ class CourtSentencingResourceIntTest : IntegrationTestBase() {
   inner class UpdateCourtAppearance {
     private val offenderNo: String = "A1234AB"
     private lateinit var courtCase: CourtCase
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingResourceIntTest.kt
```diff
@@ -5290,7 +5293,7 @@ class CourtSentencingResourceIntTest : IntegrationTestBase() {
               reportingStaff = staff,
               statusUpdateStaff = staff,
             ) {
-              offenderCharge1 = offenderCharge(resultCode1 = "1005", offenceCode = "RT88074", plea = "G")
+              offenderCharge1 = offenderCharge(resultCode1 = "1002", offenceCode = "RT88074", plea = "G")
```
andymarke, line 5296: by bad luck this 1005 is inactive so the court order is correctly being deleted, so change to an active charge code

lbennett-moj (approved):

PR #1769, opened 2026-05-19
SDIT-3808 Add accessible caseloads to returned user details

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Staff.kt
```diff
@@ -24,8 +24,6 @@ class Staff(
   @Column
   val lastName: String,
 
-  // TODO - do we need  AND INTERNET_ADDRESS_CLASS = 'EMAIL' on the sql restriction - there is only EMAIL in dev
```
karenmillermoj, line 27: I have removed this comment but then checked against preprod and it fails - so I will add a test and fix this in another PR

petergphillips (approved):
LGTM

PR #1770, opened 2026-05-19
SDIT-3808 Add roles associated with a user's caseloads 

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/Staff.kt
```diff
@@ -25,6 +26,7 @@ class Staff(
   val lastName: String,
 
   @OneToMany(mappedBy = "staff", fetch = FetchType.LAZY, cascade = [CascadeType.ALL], orphanRemoval = true)
+  @SQLRestriction("INTERNET_ADDRESS_CLASS = 'EMAIL'")
```
karenmillermoj, line 29: This feels like a bit of a hack - thinking I should subclass StaffInternetAddress and use discriminator for INTERNET_ADDRESS_CLASS

steverendell (approved):
lgtm

PR #1772, opened 2026-05-20
SDIT-3808 Pull in status for user from dba_users

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/StaffUserAccount.kt
```diff
@@ -19,23 +23,52 @@ class StaffUserAccount(
   @Column(name = "USERNAME", nullable = false)
   val username: String,
 
+  @OneToOne(fetch = FetchType.LAZY)
```
andymarke, line 26: I'm not sure a OneToOne will be lazy - best to check the SQL generated since I suspect it will always to an outer join

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/StaffStatusType.kt
```diff
@@ -0,0 +1,15 @@
+package uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa
+
+import jakarta.persistence.DiscriminatorValue
+import jakarta.persistence.Entity
+
+@Entity
```
andymarke, line 6: this *might* need to be EntityOpen - double check if it eager loading by mistake   since it is final so proxy can be created

andymarke (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/StaffStatusType.kt
karenmillermoj: We don't set up the reference codes to be entityOpen

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/StaffUserAccount.kt
karenmillermoj: Will remove if necessary in another PR

PR #1776, opened 2026-05-22
SDIT-3826 Add additional breach hearings when recall updated

andymarke:
When a recall is updated new sentences (and therefore cases) can be added to a recall and removed from a recall. We already deal with cases being removed but not new ones added.

For each cases added to a recall a breach court appearance/hearing is added.

The IDs of the court hearings are returned back to sync service to all mapping to be updated.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingResource.kt
@@ -2917,6 +2917,16 @@ data class ConvertToRecallResponse(
   val clonedCourtCases: BookingCourtCaseCloneResponse?,
 )
 
+@Schema(description = "Recall update response")
+data class UpdateRecallResponse(
+  @Schema(description = "the breach court appearance ids created")
+  val createdCourtEventIds: List<Long>,
+  @Schema(description = "the breach court appearance ids updated")
+  val updatedCourtEventIds: List<Long>,
+  @Schema(description = "the breach court appearance ids deleted")
+  val deletedCourtEventIds: List<Long>,
+)
+
 @Schema(description = "Recall convert response")
 data class SentenceIdAndAdjustmentIds(
   val sentenceId: SentenceId,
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
```diff
@@ -1349,49 +1349,20 @@ class CourtSentencingService(
...
+    sentencingAdjustmentService.convertAdjustmentsToRecallEquivalents(sentencesInRecall)
+    val adjustmentsUpdated = sentencingAdjustmentService.activateAllAdjustment(sentencesInRecall)
     convertToRecallRequest.returnToCustody.createOrUpdateBooking(bookingIds)
 
     // Create a new CourtEvent for each unique CourtCase associated with each OffenderSentence
-    val uniqueCourtCases = sentencesUpdated.mapNotNull { it.courtCase }.toSet()
+    val uniqueCourtCases = sentencesInRecall.mapNotNull { it.courtCase }.toSet()
     // Create a new CourtEvent for each unique CourtCase
     val courtEvents = uniqueCourtCases.map { courtCase ->
-      // Find all sentences associated with this court case that are being recalled
```
andymarke, line 1361: exttracted to method to be used for updates

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
```diff
@@ -1516,6 +1509,43 @@ class CourtSentencingService(
     )
   }
 
+  private fun createRecallBreachCourtAppearance(courtCase: CourtCase, recallRevocationDate: LocalDate, sentencesInRecall: List<OffenderSentence>): CourtEvent {
```
andymarke, line 1512: extracted from above

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingResource.kt
```diff
@@ -2917,6 +2917,16 @@ data class ConvertToRecallResponse(
   val clonedCourtCases: BookingCourtCaseCloneResponse?,
 )
 
+@Schema(description = "Recall update response")
```
andymarke, line 2920: Previously the PUT returned nothing - but since court appearances can now be created as a side-effect of an update we need to return these back to the sync service.

petergphillips (approved):
LGTM

PR #1780, opened 2026-05-27
SDIT-3694-non-associations-half-closed-in-nomis-cause-create-failure

steverendell:
Allow creation of a new NA when old closed NAs are not closed properly


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderExternalMovement.kt
```diff
@@ -62,7 +61,7 @@ class OffenderExternalMovement(
   @Column(name = "MOVEMENT_TIME")
   val movementTime: LocalDateTime,
 
-  @ManyToOne(fetch = FetchType.LAZY) @JoinColumns(
+  @ManyToOne(fetch = LAZY) @JoinColumns(
```
steverendell, line 64: tidying - hangover from previous PR

andymarke (approved):

PR #1781, opened 2026-05-27
SDIT-3808 Return roles as part of caseload in returned user details

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/users/UserResource.kt
```diff
@@ -152,9 +152,7 @@ data class UserAccount(
   @Schema(description = "The current active caseload on the account", example = "MDI")
   val activeCaseloadId: String? = null,
   @Schema(description = "Caseloads associated with the user", example = "['MDI','LEI']")
-  val caseloads: List<String>,
-  @Schema(description = "Roles associated with the user", example = "['USER_MANAGER','VIEW_INCIDENTS']")
-  val roles: List<String>,
+  val caseloads: List<CaseloadResponse>,
```
karenmillermoj, line 155: There is another PR to refactor/rename to use Staff (rather than User) ...coming shortly after this PR

petergphillips (approved):
LGTM

PR #1792, opened 2026-06-03
SDIT-3714 Ignore movements on different booking - must be duplicates created by merge

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/court/offender/OffenderCourtMovementsService.kt
```diff
@@ -56,8 +56,8 @@ class OffenderCourtMovementsService(
             .filter { it.offenderBooking.bookingId == bk.id }
             .map { schedule ->
               schedule.toResponse(
-                moveOut = allMovementsOut.find { it.courtScheduleOutId == schedule.id },
-                moveIn = allMovementsIn.find { it.courtScheduleOutId == schedule.id },
+                moveOut = allMovementsOut.find { it.courtScheduleOutId == schedule.id && it.offenderBooking.bookingId == bk.id },
+                moveIn = allMovementsIn.find { it.courtScheduleOutId == schedule.id && it.offenderBooking.bookingId == bk.id },
               )
```
mikehalmamoj, line 61: We had a problem when trying to link movements to schedules - these should be a 1 to 1 but in some very old bad NOMIS data there are duplicate movements created by MERGE on the wrong booking but still linking back to the same schedule. To make sure we only pick up the correct movement and ignore the duplicate we get the one matching the booking id on the schedule.

lbennett-moj (approved):

PR #1793, opened 2026-06-03
SDIT-3846 Fix bug where a charges is removed that was previously link…

andymarke:
…ed from a combined case

Two issued fixed:
1. if court_event_charge is deleted from a case and that case no longer points to that charge we correctly also delete the offender_charge. However, for a linked case there may be court_event_charge pointing at that offender_charge for the combined cases making deletion impossible. This checks if any linked case is also referencing the offender_charge before deciding to delete the offender_charge
2. LinkCaseTxn was modelled incorrectly; it assumed it would always point at CourtEventCharge. This was incorrect. You can delete the court_event_charge record, but the LINK_CASE_TXNS record still remains in NOMIS. So the fix is to make the reference nullable and to set it to null when the associated courtEventCharge is being removed.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
@@ -572,37 +572,37 @@ class CourtSentencingService(
     courtEventChargesToUpdate: List<CourtEventChargeRequest>,
     courtEvent: CourtEvent,
   ) {
-    val originalList = courtEvent.courtEventCharges
-    val newChargeList = mutableListOf<CourtEventCharge>()
-    courtEventChargesToUpdate.map { cecRequest ->
-      val resultCode = cecRequest.resultCode1?.let { lookupOffenceResultCode(it) }
-      courtEvent.courtEventCharges.firstOrNull { it.id.offenderCharge.id == cecRequest.offenderChargeId }
-        ?.let { existingCourtEventCharge ->
-          existingCourtEventCharge.resultCode1 = resultCode
-          existingCourtEventCharge.resultCode1Indicator = resultCode?.dispositionCode
-          newChargeList.add(existingCourtEventCharge)
-        } ?: let {
-        getOffenderCharge(cecRequest.offenderChargeId).let { offenderCharge ->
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
```diff
@@ -572,37 +572,37 @@ class CourtSentencingService(
...
-              offenceDate = offenderCharge.offenceDate,
-              offenceEndDate = offenderCharge.offenceEndDate,
-              mostSeriousFlag = offenderCharge.mostSeriousFlag,
-              resultCode1 = resultCode,
-              resultCode1Indicator = resultCode?.dispositionCode,
+    val chargesToRemove = courtEvent.courtEventCharges.filter { charge ->
+      charge.id.offenderCharge.id !in courtEventChargesToUpdate.map { it.offenderChargeId }
+    }
+    // the linked transaction remains but no longer points at the court event charge, just the court event and offender charge
+    chargesToRemove.forEach { it.linkedCaseTransaction?.courtEventCharge = null }
```
andymarke, line 579: this is the only significant change here ,  clear the link in JPA object else it will be a transient reference that can never be persisted

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/LinkCaseTxn.kt
```diff
@@ -37,14 +37,14 @@ data class LinkCaseTxn(
...
-  @OneToOne(fetch = FetchType.LAZY, optional = false)
+  @OneToOne(fetch = FetchType.LAZY, optional = true)
   @JoinColumns(
     value = [
       JoinColumn(name = "OFFENDER_CHARGE_ID", referencedColumnName = "OFFENDER_CHARGE_ID", nullable = true, insertable = false, updatable = false),
       JoinColumn(name = "EVENT_ID", referencedColumnName = "EVENT_ID", nullable = true, insertable = false, updatable = false),
     ],
   )
-  var courtEventCharge: CourtEventCharge,
+  var courtEventCharge: CourtEventCharge?,
```
andymarke, line 48: The LINK_CASE_TXNS record is kept in NOMIS even if the COURT_EVENT_CHARGE has been deleted, so this has to be nullable

lbennett-moj (approved):

PR #1794, opened 2026-06-03
SDIT-3846 Fix bug where a charges is removed that was previously link…

andymarke:
…ed from a combined case

Fix issue where mappings was being deleted since endpoint was incorrectly returning the linked charge as deleted.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
@@ -724,10 +724,8 @@ class CourtSentencingService(
     courtCase: CourtCase,
     offenderNo: String,
     eventId: Long,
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
```diff
@@ -724,10 +724,8 @@ class CourtSentencingService(
     courtCase: CourtCase,
     offenderNo: String,
     eventId: Long,
-  ): List<OffenderCharge> = courtCase.getOffenderChargesNotAssociatedWithCourtAppearances().also { orphanedOffenderCharges ->
+  ): List<OffenderCharge> = courtCase.getOffenderChargesNotAssociatedWithCaseOrLinkedCasesCourtAppearances().also { orphanedOffenderCharges ->
     orphanedOffenderCharges
-      // exclude where a linked case source still points to this charge
```
andymarke, line 729: moved this down in the above method - else we were incorrectly return the charge as deleted when in fact it wasn't

lbennett-moj (approved):

PR #1800, opened 2026-06-05
SDIT-3710 Add special audit module for court scheduler sync updates

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/court/schedule/CourtScheduleService.kt
```diff
@@ -41,6 +42,7 @@ class CourtScheduleService(
   }
 
   @Transactional
+  @Audit(auditModule = "DPS_SYNCHRONISATION_COURT_SCHEDULER")
```
mikehalmamoj, line 45: We need a dedicated audit module for court scheduler because we don't want to ignore court event updates syncd from to NOMIS from court sentencing.

andymarke (approved):

PR #1801, opened 2026-06-05
SDIT-3710 Use a non-standard audit module name

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/court/schedule/CourtScheduleService.kt
```diff
@@ -42,7 +42,7 @@ class CourtScheduleService(
   }
 
   @Transactional
-  @Audit(auditModule = "DPS_SYNCHRONISATION_COURT_SCHEDULER")
+  @Audit(auditModule = "DPS_COURT_SCHEDULER_SYNCHRONISATION")
```
mikehalmamoj, line 45: Deliberately used a non-standard audit module name (e.g. does not start with `DPS_SYNCHRONISATION`) so that we don't interfere with the NOMIS to DPS RaS sync.

andymarke (approved):

steverendell (commented):
lgtm

PR #1802, opened 2026-06-05
SDIT-3865 Changed orphaned offender charges detection

andymarke:
Fix bug where a charges is removed that was previously linked from a combined case that itself was previously later unlinked but the charge is still referenced by both cases


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
```diff
@@ -2096,6 +2096,19 @@ class CourtSentencingService(
       },
     )
   }
+
```
andymarke, line 2099: moved into class so can access `courtEventChargeRepository`.

The change is previously it would only look at linked cases by now uses SQL to look at all cases so included cases that were linked but are no longer linked.

mikehalmamoj (approved):

PR #1825, opened 2026-06-16
SDIT-3885: 🎨 Tidy use of event date and start time

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/CourtEvent.kt
```diff
@@ -146,12 +155,7 @@ class CourtEvent(
   }
 
   fun isLatestAppearance(): Boolean = courtCase?.let {
-    this == courtCase!!.courtEvents.sortedBy { event ->
```
petergphillips, line 149: intellij suggestion that a `sortedBy` followed by a `last` can be simplified to `maxByOrNull`

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/court/schedule/CourtScheduleService.kt
```diff
@@ -135,7 +135,7 @@ class CourtScheduleService(
     bookingId = offenderBooking.bookingId,
     eventId = id,
     eventDate = eventDate,
-    startTime = eventDate.atTime(startTime.toLocalTime()),
```
petergphillips, line 138: simplify all usage to call new method

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
```diff
@@ -671,7 +671,7 @@ class CourtSentencingService(
   }
 
   private fun updateCourtIfNecessary(courtCase: CourtCase) {
-    courtCase.courtEvents.maxWithOrNull(compareBy<CourtEvent> { it.eventDate }.thenBy { it.startTime })
```
petergphillips, line 674: simplified, but not sure it was correct previously if the `startTime` was set to the wrong date 🤯

andymarke (approved):

PR #1826, opened 2026-06-16
SDIT-3885: 🐛 Fix use of transaction event date and time

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/GeneralLedgerTransaction.kt
```diff
@@ -65,7 +65,7 @@ data class GeneralLedgerTransaction(
   val entryDate: LocalDate,
 
   @Column(name = "TXN_ENTRY_TIME", nullable = false)
-  val entryTime: LocalTime,
+  val entryTime: LocalDateTime,
```
petergphillips, line 68: this would have been wrong if we started inserting as it would have set the date to the correct date. This would have then meant that the datetime corrections package would have had to correct every single one every time it ran.

mikehalmamoj (approved):

PR #1827, opened 2026-06-17
SDIT-3885: 🎨 Ensure access to time is through new getter

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/GeneralLedgerTransaction.kt
```diff
@@ -65,7 +65,7 @@ data class GeneralLedgerTransaction(
   val entryDate: LocalDate,
 
   @Column(name = "TXN_ENTRY_TIME", nullable = false)
-  val entryTime: LocalDateTime,
+  private val entryTime: LocalDateTime,
```
petergphillips, line 68: `private` will stop people accessing it outside the new getter

mikehalmamoj (approved):

PR #1831, opened 2026-06-19
SDIT-3885: 🎨 Switch to time getters and setters

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderAppointment.kt
```diff
@@ -53,4 +54,19 @@ class OffenderAppointment(
...
+) {
+  /**
+   * Return the appointment end date with the time portion set to the appointment end time. Under some circumstances
+   * (and until corrected by TAG_DATETIME_CORRECTIONS) the date portion of the appointment end time may be different,
+   * so need to combine the two to ensure we get the correct date and time.
+   *
+   * @return The combined LocalDateTime representing the appointment end date and time.
+   */
+  fun getAppointmentEndDateAndTime(): LocalDateTime? = endTime?.let {
+    eventDate!!.atTime(it.toLocalTime())
```
petergphillips, line 66: existing usage always seems to be if the time is set then date must be set

steverendell (approved):
lgtm

PR #1832, opened 2026-06-19
SDIT-3845 Return latest booking flag for court appearances

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/court/schedule/CourtScheduleOut.kt
```diff
@@ -10,6 +10,9 @@ data class CourtScheduleOut(
   @Schema(description = "Booking ID")
   val bookingId: Long,
 
+  @Schema(description = "Is this the latest booking for the offender?")
+  val latetstBooking: Boolean,
+
```
andymarke, line 15: obviously it doesn't matter but I think I tend to return the booking sequence and then convert to the currentTerm in the sync service .  Not sure it matters if we should be consistent - too small to worry about

petergphillips (approved):
LGTM

PR #1839, opened 2026-06-24
SDIT-3977 Basic endpoint and sub-class JPA mapping

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/agency/AgencyResource.kt
```diff
@@ -0,0 +1,176 @@
...
+      ),
+    ],
+  )
+  fun getAgency(
+    @PathVariable
+    @Schema(description = "Agency id (aka agencyId)", example = "WWI")
+    agencyId: String,
+  ) = agencyService.getAgency(agencyId)
+
+  @GetMapping("/agency-location/{agencyId}")
```
andymarke, line 116: Silly test - but just proves we can still treat a Prison as a AgencyLocation instance

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/taps/OffenderTapsResourceIntTest.kt
```diff
@@ -77,7 +77,7 @@ class OffenderTapsResourceIntTest(
     }
     offenderRepository.deleteAll()
     if (this::agencyLocation.isInitialized) {
-      agencyLocationRepository.delete(agencyLocation)
+      agencyLocationRepository.deleteById(agencyLocation.id)
     }
```
andymarke, line 81: Can't pass a Prison into agencyLocationRepository any more for save or delete - so simple cheat - just deleted by id

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/taps/TapRepositoryIntTest.kt
```diff
@@ -208,7 +216,7 @@ class TapRepositoryIntTest(
         corporate("Kwikfit") {
           corporateAddress = address()
         }
-        agencyLocation("LEI", "HMP Leeds") {
```
andymarke, line 211: LEI already exists from seed data so this actually trying to update via AgencyLocationRepository which will no longer work.

Long term we will separate builders in to correct sub-class - but makes sense to change to a prison that does't exist

mikehalmamoj (approved):

PR #1840, opened 2026-06-25
SDIT-3710 Don't return court schedule IDs that don't exist

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderCourtMovementOut.kt
```diff
@@ -22,7 +20,6 @@ class OffenderCourtMovementOut(
   commentText: String? = null,
 
   @Column(name = "EVENT_ID")
-  @NotFound(action = NotFoundAction.IGNORE)
```
mikehalmamoj, line 25: This only works for entity relationships - does nothing here.

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/court/movement/CourtMovementService.kt
```diff
@@ -58,7 +60,7 @@ class CourtMovementService(
     bookingId = id.offenderBooking.bookingId,
     sequence = id.sequence,
     eventId = courtScheduleOutId,
-    courtScheduleOutId = courtScheduleOutId,
+    courtScheduleOutId = courtScheduleOutId?.takeIf { courtEventRepository.existsById(it) },
```
mikehalmamoj, line 63: Because NOMIS has no referential integrity it's possible to have a link to a court schedule that doesn't exist. Therefore we'll have to check for ourselves if it's really there.

andymarke (approved):

PR #1842, opened 2026-06-26
SDIT-3977 Added district mappings

andymarke:
Checked live data but not sure if these are correct reference tables, it might be just AREAS table instead - but this will not break anything


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AgencyLocation.kt
```diff
@@ -94,23 +98,53 @@ class AgencyLocation(
...
+  active: Boolean,
 
   @ManyToOne(fetch = LAZY)
   @JoinColumnsOrFormulas(
     value = [
       JoinColumnOrFormula(
         formula = JoinFormula(
-          value = "'" + GeographicType.GEOGRAPHIC + "'",
+          value = "'" + AreaType.AREA + "'",
           referencedColumnName = "domain",
```
andymarke, line 144: Had it the wrong way round

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/AgencyLocation.kt
```diff
@@ -79,6 +92,38 @@ class AgencyLocationBuilder(
     .let { repository.save(it) }
     .also { agencyLocation = it }
```
andymarke, line 94: Keeping all 3 variations in the same builder for now - might separate in future - but given they all go to same table might make sense to stay in same class

petergphillips (approved):
LGTM

PR #1843, opened 2026-06-26
SDIT-3977 Corrected NotFound on district

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AgencyLocation.kt
```diff
@@ -44,8 +44,6 @@ class AgencyLocation(
   @Column(name = "DESCRIPTION")
   val description: String,
 
-  // FD not in reference data, so ignore if not found
```
andymarke, line 47: Added to the wrong column

mikehalmamoj (approved):

PR #1844, opened 2026-06-26
SDIT- 3977 agency location mapping

andymarke:
A lot repetition since currently have 3 models and 3 endpoints which will might eventually reduce to 2 - so for now being cautious not to break the JPA model


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AgencyLocation.kt
```diff
@@ -61,7 +62,7 @@ class AgencyLocation(
       ),
     ],
   )
-  val type: AgencyLocationType? = null,
+  val type: AgencyLocationType = PRISON_TYPE,
```
andymarke, line 66: not nullable  - but there is a ton of unit tests that don't pass value so have to default

karenmillermoj (approved):
lgtm

PR #1845, opened 2026-06-26
Sdit 3977 agency location mapping

andymarke:
Added:
* UPDATED_ALLOWED_FLAG
* CONTACT_NAME
* JURISDICTION_CODE

Might be a reason at this point to create a Court subclass - but confusingly the "prison" location "OUT" has a JURISDICTION_CODE of Crown Court - so if we find more Court only fields we will drop courtType from all other iagency types and just add to Courts and see what hacks we can use to ignore the OUT court type

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/NomisData.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/NomisData.kt
@@ -493,6 +493,7 @@ class NomisData(
     active: Boolean,
     districtCode: String?,
     deactivationDate: LocalDate?,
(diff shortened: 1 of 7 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/NomisData.kt
```diff
@@ -592,7 +610,6 @@ class NomisData(
 
 @NomisDataDslMarker
 interface NomisDataDsl {
-  @OffenderDslMarker
```
andymarke, line 595: all redundancy markers giving IntelliJ warnings

steverendell (approved):
lgtm

PR #1848, opened 2026-06-29
SDIT-3977 agency location mapping

andymarke:
Fields mapped:

* DISABILITY_ACCESS_CODE
* SUB_AREA_CODE
* AREA_CODE

Includes skeleton of Area which will also be migrated to DPS

Some agencies belong a sub-area within an area.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/TypeOfArea.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/TypeOfArea.kt
@@ -1,0 +1,13 @@
+package uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa
+
+import jakarta.persistence.DiscriminatorValue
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/TypeOfArea.kt
```diff
@@ -0,0 +1,13 @@
+package uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa
+
+import jakarta.persistence.DiscriminatorValue
+import jakarta.persistence.Entity
+
+@Entity
+@DiscriminatorValue(TypeOfArea.AREA_TYPE)
+class TypeOfArea(code: String, description: String) : ReferenceCode(AREA_TYPE, code, description) {
```
andymarke, line 8: There was already a AreaType that models the AREA reference codes (not to be confused with the AREA dynamic data)

mikehalmamoj (approved):

PR #1850, opened 2026-06-29

andymarke:
* region

An area can have a parent region or can be top level (e.g. Northern Ireland)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AgencyLocation.kt
```diff
@@ -127,6 +127,11 @@ class AgencyLocation(
   @JoinColumn(name = "AREA_CODE", referencedColumnName = "AREA_CODE", nullable = true)
   var area: Area? = null,
 
+  // NOMIS form maps this to REFERENCE_CODES <GEOGRAPHIC> - but data is from AREAS
```
andymarke, line 130: So weird that the NOMIS screen is just wrong; I guess it is always set via a script rather than the UI

mikehalmamoj (approved):

PR #1852, opened 2026-06-30
SDIT-3998 Only return account details for those accounts with non-zer…

karenmillermoj:
…o balance or holdBalance


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/finance/PrisonerBalanceResourceIntTest.kt
```diff
@@ -316,24 +316,18 @@ class PrisonerBalanceResourceIntTest : IntegrationTestBase() {
       with(balance) {
         assertThat(rootOffenderId).isEqualTo(id2)
         assertThat(prisonNumber).isEqualTo("B2345CD")
-        assertThat(accounts.size).isEqualTo(3)
+        assertThat(accounts.size).isEqualTo(2)
```
karenmillermoj, line 319: This is now only 2 as it no longer returns the account with null/zero balances

mikehalmamoj (approved):

PR #1854, opened 2026-06-30
SDIT-3986 Fix bug where we update court event IN to same status as court event OUT

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/court/schedule/CourtScheduleService.kt
```diff
@@ -89,7 +89,7 @@ class CourtScheduleService(
         this.parentEventId = savedScheduleOut.id
         this.setEventDateAndTime(request.startTime.with(DEFAULT_IN_TIME))
         this.courtEventType = courtEventType
-        this.eventStatus = eventStatus
+        this.eventStatus = returnEventStatus
```
mikehalmamoj, line 92: Stop updating the inbound movement with the outbound status

karenmillermoj (approved):
lgtm

PR #1857, opened 2026-07-01

andymarke:
* AGENCY_LOCATION_AUTHORITIES


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AgencyLocationAuthority.kt
```diff
@@ -0,0 +1,83 @@
...
+        referencedColumnName = "AGY_LOC_ID",
+        insertable = false,
+        updatable = false,
+      ),
+    ],
+  )
+  val agencyLocation: AgencyLocation,
+
+  @Column(name = "LOCAL_AUTHORITY_CODE")
+  val localAuthorityCode: String,
```
andymarke, line 36: Interestingly it wouldn't allow me to use the `JoinColumnsOrFormulas` actual LocalAuthorityType as part of composite primary key due to the formulas - so have to map this twice

steverendell (approved):
lgtm

PR #1858, opened 2026-07-01

andymarke:
* Address
* Address Phones


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/AgencyLocation.kt
```diff
@@ -47,9 +50,6 @@ interface AgencyLocationDsl {
     comment: String? = null,
     startDate: String? = null,
     endDate: String? = null,
-    isServices: Boolean = false,
```
andymarke, line 50: never used for agency addresses so removed from DSL

mikehalmamoj (approved):

PR #1860, opened 2026-07-01

andymarke:
re-mapped district_code to Area from Reference_Codes. This simplifies AgencyLocation hierarchy since all sub-types are identical so can be removed in next PR


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/AgencyLocation.kt
```diff
@@ -214,33 +221,12 @@ class Prison(
   longDescription: String?,
   payrollRegion: PayrollRegionType?,
 
-  // FD (for OUT and TRN) not in reference data, so ignore if not found
```
andymarke, line 217: This is a bit like MOVEMENT_REASONS and REFERENCE_CODES - the codes exist on both tables, so switch to AREAS table for now - since I still can't work how this column is populated in NOMIS

mikehalmamoj (approved):

PR #1864, opened 2026-07-02
SDIT-3977 AgencyLocation to not use inheritance

andymarke:
All types of agencies have same behaviour; a minor difference is only courts have a court type (though NOMIS does not enforce this) - so by removing hierarchy we reduce code bloat

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/agency/AgencyService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/agency/AgencyService.kt
@@ -5,36 +5,26 @@ import org.springframework.stereotype.Service
 import org.springframework.transaction.annotation.Transactional
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.data.NotFoundException
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.data.toCodeDescription
-import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.Agency
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.AgencyLocation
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/AgencyLocationRepository.kt
```diff
@@ -5,19 +5,11 @@ import org.springframework.data.jpa.repository.JpaSpecificationExecutor
 import org.springframework.stereotype.Repository
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.AgencyLocation
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.AgencyLocationType
-import java.util.Optional
 
 @Repository
 interface AgencyLocationRepository :
   JpaRepository<AgencyLocation, String>,
   JpaSpecificationExecutor<AgencyLocation> {
-  fun findByIdAndDeactivationDateIsNull(id: String): Optional<AgencyLocation>
```
andymarke, line 14: not used

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/agency/AgencyService.kt
```diff
@@ -94,6 +38,8 @@ fun AgencyLocation.toAgencyLocationResponse() = AgencyLocationResponse(
   addresses = this.toAgencyAddresses(),
   phones = this.toPhoneNumbers(),
   emailAddresses = this.toEmailAddresses(),
+  district = this.district?.toCodeDescription(),
+  courtType = this.courtType?.toCodeDescription(),
 )
```
andymarke, line 43: accidentally missing

lbennett-moj (approved):

PR #1866, opened 2026-07-03
SDIT-4016 Responsible prison should be the location of the prisoner

mikehalmamoj:
The location being where the prisoner was when the movement took place - which could be OUT, or a prison on the previous booking

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderExternalMovementRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderExternalMovementRepository.kt
@@ -3,10 +3,8 @@ package uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.repository
 import org.springframework.data.jpa.repository.Query
 import org.springframework.data.repository.CrudRepository
 import org.springframework.stereotype.Repository
-import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.AgencyLocation
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.OffenderExternalMovement
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.OffenderExternalMovementId
-import java.time.LocalDateTime
 
 @Repository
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderExternalMovementRepository.kt
```diff
@@ -98,13 +96,5 @@ interface OffenderExternalMovementRepository : CrudRepository<OffenderExternalMo
   fun countOffenderScheduledIn(offender: String): Long
 
   @Suppress("ktlint:standard:function-naming")
-  fun findAllById_OffenderBooking_BookingId(bookingId: Long): List<OffenderExternalMovement>
-
-  fun findPrisonAt(time: LocalDateTime, bookingId: Long): AgencyLocation? {
```
mikehalmamoj, line 103: Move to a service because it now needs 2 repositories

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/service/ExternalMovementService.kt
```diff
@@ -0,0 +1,26 @@
...
+  fun findPrisonAt(time: LocalDateTime, offenderNo: String): AgencyLocation? {
+    val movements = externalMovementRepository.findAllById_OffenderBooking_Offender_NomsId(offenderNo)
+      .filter { it.movementReason.id.type == "ADM" || it.movementReason.id.type == "REL" }
+    return movements
+      .filter { it.getMovementDateAndTime() < time }
+      .maxByOrNull { it.getMovementDateAndTime() }
+      ?.takeIf { it.movementReason.id.type == "ADM" }
+      ?.toAgency
+      ?: agencyLocationRepository.findByIdOrNull("OUT")
+  }
```
mikehalmamoj, line 25: The rules for determining the responsible prison.
1. if the previous admission / release is an admission, use that prison
2. if the previous admission / release is a release, use OUT
3. for anything else, assume OUT

steverendell (approved):
lgtm

PR #1876, opened 2026-07-10
SDIT-4018: 🐛 Change audit module to be nullable

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incentives/IncentiveResponse.kt
```diff
@@ -14,22 +14,19 @@ data class IncentiveResponse(
...
   @Schema(description = "IEP level code and description", required = true)
   val iepLevel: CodeDescription,
-  @Schema(description = "User id of user creating prisoner incentive level", required = false)
+  @Schema(description = "User id of user creating prisoner incentive level")
   val userId: String? = null,
   @Schema(description = "Is this IEP the current IEP for the booking?", required = true)
   val currentIep: Boolean,
-  @Schema(
-    description = "The NOMIS module that created this IEP",
-    required = true,
```
petergphillips, line 31: it is not required 🤦

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/incentives/IncentiveResponse.kt
```diff
@@ -14,22 +14,19 @@ data class IncentiveResponse(
   val bookingId: Long,
   @Schema(description = "The sequence of the incentive within this booking", required = true)
   val incentiveSequence: Long,
-  @Schema(description = "Comment for Incentive level", required = false)
```
petergphillips, line 17: default is `false` anyway so no need to provide

mikehalmamoj (approved):

PR #1877, opened 2026-07-10
SDIT-4030 Change null outcome repair

andymarke:
Will now use latest court event outcome and also call the synchronous version of imprisonment status store procedure


petergphillips (approved):
LGTM

PR #1879, opened 2026-07-10
SDIT-4032 Handle race to delete schedule

mikehalmamoj:
We can receive concurrent events to delete a schedule - one from an application update, one from a direct delete request - and we had an error where one pod deleted a schedule another pod thought existed ending with a stale object exception and a rejected delete. While nothing went wrong, we ended having to investigate a DLQ message, and this is not idempotent behaviour from the delete endpoint.

So to fix we always lock the row before deleting - then it doesn't matter which pod wins the race.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderTapScheduleOutRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderTapScheduleOutRepository.kt
@@ -2,7 +2,10 @@
 
 package uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.repository
 
+import jakarta.persistence.LockModeType
 import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Lock
+import org.springframework.data.jpa.repository.Query
 import org.springframework.stereotype.Repository
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.OffenderTapScheduleOut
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/taps/schedule/TapScheduleService.kt
```diff
@@ -140,7 +140,7 @@ class TapScheduleService(
 
   @Transactional
   fun deleteTapScheduleOut(offenderNo: String, eventId: Long) {
-    scheduleOutRepository.findByIdOrNull(eventId)
+    scheduleOutRepository.findByEventIdOrNullForUpdate(eventId)
```
mikehalmamoj, line 143: It would have been nice just to call `deleteById()` but we have to load the schedule first to perform some validation.

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderTapScheduleOutRepository.kt
```diff
@@ -11,4 +14,8 @@ interface OffenderTapScheduleOutRepository : JpaRepository<OffenderTapScheduleOu
   fun findByEventIdAndOffenderBooking_Offender_NomsId(eventId: Long, offenderNo: String): OffenderTapScheduleOut?
 
   fun countByOffenderBooking_Offender_NomsId_AndTapApplicationIsNotNull(offenderNo: String): Long
+
+  @Lock(LockModeType.PESSIMISTIC_WRITE)
+  @Query(value = "select tso from OffenderTapScheduleOut tso where (tso.eventId = :eventId)")
```
karenmillermoj, line 19: do you need to set a timeout or are you counting on the lock eventually being released (which I suppose it should) ?

karenmillermoj (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderTapScheduleOutRepository.kt
mikehalmamoj: These aren't for a lock in the UI, so any lock should only last for milliseconds.

PR #1883, opened 2026-07-20
SDIT-4051 Adjudication room lookup fix

andymarke:
When looking up a room for the dummy hearing and no location with usage type of 'ADJU' found then full back to looking for any previous room that has been used for referrals

FDI has changed their room types so dummy hearings can no longer be created.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationsHearingResultsResourceIntTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationsHearingResultsResourceIntTest.kt
@@ -409,6 +409,7 @@ class AdjudicationsHearingResultsResourceIntTest : IntegrationTestBase() {
     private val existingAdjudicationNumber = 123456L
     private lateinit var existingCharge: AdjudicationIncidentCharge
     private lateinit var existingSecondCharge: AdjudicationIncidentCharge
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/adjudications/AdjudicationsHearingResultsResourceIntTest.kt
```diff
@@ -558,6 +560,31 @@ class AdjudicationsHearingResultsResourceIntTest : IntegrationTestBase() {
...
+          .expectStatus().isOk
+
+        repository.runInTransaction {
+          val adjudicationHearing = hearingRepository.findByAdjudicationNumber(existingAdjudicationNumber)[1]
+          assertThat(adjudicationHearing.comment).isEqualTo("DPS_REFERRAL_PLACEHOLDER-${existingThirdCharge.id.chargeSequence}")
+          assertThat(adjudicationHearing.hearingDateTime).isEqualTo(LocalDateTime.now().with(LocalTime.MIDNIGHT))
+          assertThat(adjudicationHearing.hearingResults[0].findingType.code).isEqualTo("REF_POLICE")
+
+          // now switch back to Adjudication type
+          adjudicationHearing.agencyInternalLocation?.locationType = "ADJU"
```
andymarke, line 587: bit of risk hear if test fails since reference data would be wrong state for later tests - a risk worth taking to remove setup complexity

petergphillips (approved):
LGTM

PR #1886, opened 2026-07-20
SDIT-3535: 🐛 Cope with bad data by not mapping booking

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderBelief.kt
```diff
@@ -26,9 +26,8 @@ data class OffenderBelief(
   @Column(name = "BELIEF_ID", nullable = false)
   var beliefId: Long = 0,
 
-  @ManyToOne
-  @JoinColumn(name = "OFFENDER_BOOK_ID", nullable = false)
-  val booking: OffenderBooking,
+  @Column(name = "OFFENDER_BOOK_ID", nullable = false)
+  val booking: Long,
```
steverendell, line 30: a tiny comment , but should be called bookingId?

steverendell (approved):
lgtm

PR #1890, opened 2026-07-21
SDIT-4047 Sentence create booking locking

andymarke:
Two inter-linked issues here:
* The locking via a JOIN between OFFENDER_BOOKINGS and OFFENDER_CASES does not appear to work; so now acquire lock on booking and case
* The call to imprisonment status stored procedure appeared to be executed on a 2nd connection; causing the rollback to fail for main transaction when stored procedure fails.

Call store procedure in contect of a JDBCTemplate (that doesn't appear to get a new conection but uses the one associated with the current transaction)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
@@ -6,6 +6,7 @@ import org.springframework.beans.factory.annotation.Value
 import org.springframework.data.domain.Page
 import org.springframework.data.domain.Pageable
 import org.springframework.data.repository.findByIdOrNull
+import org.springframework.jdbc.core.JdbcTemplate
 import org.springframework.stereotype.Service
 import org.springframework.transaction.annotation.Transactional
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.audit.AuditRepository
@@ -111,6 +112,7 @@ class CourtSentencingService(
   private val sentencingAdjustmentService: SentencingAdjustmentService,
   private val linkCaseTxnRepository: LinkCaseTxnRepository,
   private val auditRepository: AuditRepository,
+  private val jdbcTemplate: JdbcTemplate,
   @Value("\${spring.datasource.username}") val datasourceUsername: String,
 ) {
   private companion object {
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
```diff
@@ -902,70 +913,72 @@ class CourtSentencingService(
     return latestWithResultCode == courtEventCharge
   }
 
-  fun createSentence(offenderNo: String, caseId: Long, request: CreateSentenceRequest) = findCourtCaseWithOffenderBookingLock(caseId = caseId, offenderNo = offenderNo).let { case ->
+  fun createSentence(offenderNo: String, caseId: Long, request: CreateSentenceRequest): CreateSentenceResponse {
+    findCourtCaseWithLock(id = caseId, offenderNo = offenderNo).let { case ->
+      val offenderBooking = findOffenderBookingWithLock(case.offenderBooking.bookingId)
```
andymarke, line 918: Obtain lock on OFFENDER_BOOKINGS as well since cases can be created in parallel across an entire booking by DPS

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
```diff
@@ -902,70 +913,72 @@ class CourtSentencingService(
...
+          courtEvent = findCourtAppearance(offenderNo = offenderNo, id = request.eventId),
+        )
+          ?: throw BadDataException("Court order not found for booking ${offenderBooking.bookingId} and court event ${request.eventId}"),
+        // this is the sentence sequence this sentence is consecutive to
+        consecSequence = request.consecutiveToSentenceSeq?.toInt(),
+        lineSequence = offenderSentenceRepository.getNextLineSequence(offenderBooking).toInt(),
+        statusUpdateReason = "A",
+      )
 
-    val offenderBooking = case.offenderBooking
```
andymarke, line 907: No change here - just diff bad due to the whole function moving to the right

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/repository/StoredProcedureRepository.kt
```diff
@@ -54,13 +55,14 @@ class StoredProcedureRepositoryOracle(
...
+    jdbcTemplate: JdbcTemplate,
     bookingId: Long,
     changeType: String,
   ) {
     val params = MapSqlParameterSource()
       .addValue("p_offender_book_id", bookingId)
       .addValue("p_change_type", changeType)
-    updateStatusAndMainOffence.execute(params)
+    UpdateStatusAndMainOffence(jdbcTemplate).execute(params)
   }
```
andymarke, line 66: Not sure if this has any performance implications given  that this is initialise for each request.

mikehalmamoj (approved):

PR #1893, opened 2026-07-22
SDIT-4052 Take court event prison from the offender booking

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/court/schedule/CourtScheduleService.kt
```diff
@@ -91,7 +90,7 @@ class CourtScheduleService(
         this.courtEventType = courtEventType
         this.eventStatus = returnEventStatus
         this.commentText = request.comment
-        this.court = prison
+        this.court = offenderBooking.location
```
mikehalmamoj, line 93: DPS are no longer telling us which prison to return the prisoner to after a court appearance.

The scheduled court movement IN is currently created by NOMIS when the prisoner moves OUT to court - but we still maintain the schedule in NOMIS if anything changes in DPS (and will be creating these records once DPS handle the movements). The prisoner  will always be coming back to whichever prison they were in at the time (unless we end up in a court transfer situation, which isn't this scenario).

andymarke (approved):

PR #1896, opened 2026-07-22
SDIT-4058 removing the religion call as the religions call provides a…

richardpopple:
…ll the necessary information


petergphillips (approved):
LGTM

PR #1904, opened 2026-07-24
SDIT-4069 Ensure allow for empty gl_transaction entries so pick up tr…

karenmillermoj:
…ansactionTimestamp from offender_transactions entry


petergphillips (approved):
LGTM

PR #1905, opened 2026-07-24
SDIT-4062 Imprisonment Status / Mapping table and data

andymarke:
Data taken from prod so alos had to update the offence result codes and clac types; which then means I had to update to test data to make it valid or invalid as appropiate


petergphillips (approved):
LGTM

PR #1906, opened 2026-07-24
SDIT-4072 Reject case cloning for future dated appearance

andymarke:
The issue is if a warrant with a next appearance is added but on a case on an old booking we react to the court appearance insert by cloning the court case. Since we might also get an event for next appearance at same time we can then clone the case twice causing chaos.

This will reject the clone for a next appearance and rely on sync retrying (await parent style)

NB there is potential for a extreme rare bug if DPS court appearance on old booking was amended to add a next court appearance while still being on old booking. This would result on message going on DLQ for manual intervention.


petergphillips (approved):
LGTM

PR #1907, opened 2026-07-27
SDIT-4062 OffenderImprisonmentStatus JPA mapping

andymarke:
Also fixed OffenderBookings DSL markers


src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/builders/AdjudicationParty.kt
```diff
@@ -24,7 +24,7 @@ import java.time.LocalDateTime
 @DslMarker
 annotation class AdjudicationPartyDslMarker
```
andymarke, line 26: all these should use own Marker not Nomis top level one

mikehalmamoj (approved):

PR #1916, opened 2026-07-29

andymarke:
Cleaned up SQL where not used

* TODO locking strategy
* Switch to use this service from Stored Procedure


karenmillermoj (approved):
lgtm

PR #1917, opened 2026-07-29

andymarke:
Now lock booking record to prevent service being called concurrently for same booking.

Also try to acquire lock for each of these records that can be updated:

* OFFENDER_IMPRISON_STATUSES
* OFFENDER_CHARGES
* COURT_EVENT_CHARGES

This should prevent a NOMIS lock from hanging this service indefinitely

* Switch to use this service from Stored Procedure


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderChargeRepository.kt
```diff
@@ -17,4 +17,9 @@ interface OffenderChargeRepository : JpaRepository<OffenderCharge, Long> {
   @QueryHints(value = [QueryHint(name = "jakarta.persistence.lock.timeout", value = "2000")])
   @Query("SELECT c FROM OffenderCharge c WHERE c.id = :id")
   fun findByIdOrNullForUpdate(id: Long): OffenderCharge?
+
+  @Lock(LockModeType.PESSIMISTIC_WRITE)
+  @QueryHints(value = [QueryHint(name = "jakarta.persistence.lock.timeout", value = "1000")])
+  @Query("SELECT c FROM OffenderCharge c WHERE c.id = :id")
+  fun findByIdWaitForLock(id: Long): OffenderCharge
```
andymarke, line 24: similar to above - but expect record to be there and use smallest wait time

petergphillips (approved):
LGTM

PR #1918, opened 2026-07-29

andymarke:
Add feature switch (profile) so we can switch to either version

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
@@ -6,7 +6,6 @@ import org.springframework.beans.factory.annotation.Value
 import org.springframework.data.domain.Page
 import org.springframework.data.domain.Pageable
 import org.springframework.data.repository.findByIdOrNull
-import org.springframework.jdbc.core.JdbcTemplate
 import org.springframework.stereotype.Service
 import org.springframework.transaction.annotation.Transactional
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.audit.AuditRepository
@@ -73,7 +72,6 @@ import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.repository.ReferenceCod
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.repository.SentenceCalculationTypeRepository
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.repository.StaffUserAccountRepository
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.repository.CourtOrderInsertRepository
-import uk.gov.justice.digital.hmpps.nomisprisonerapi.repository.StoredProcedureRepository
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.repository.storedprocs.ImprisonmentStatusChangeType
 import uk.gov.justice.digital.hmpps.nomisprisonerapi.sentencingadjustments.SentencingAdjustmentService
 import java.time.LocalDate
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/ImprisonmentStatusService.kt
```diff
@@ -34,10 +37,9 @@ class ImprisonmentStatusService(
     }
   }
 
-  fun recalculateImprisonmentStatusAndMainOffence(offenderNo: String, reason: ChangeType): ImprisonmentStatusAndMainOffence {
-    val booking = offenderBookingRepository.findLatestBookingIdByOffenderNomsId(offenderNo)
-      ?.let { offenderBookingRepository.findByIdOrNullForUpdate(it) }
-      ?: throw NotFoundException("No booking found for offenderNo $offenderNo")
+  fun recalculateImprisonmentStatusAndMainOffence(bookingId: Long, reason: ChangeType): ImprisonmentStatusAndMainOffence {
```
andymarke, line 40: We always have bookingId that is more accurate since you can change sentences etc on previous bookings

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
```diff
@@ -105,15 +103,14 @@ class CourtSentencingService(
...
-  private val storedProcedureRepository: StoredProcedureRepository,
   private val sentenceCalculationTypeRepository: SentenceCalculationTypeRepository,
   private val mergeTransactionRepository: MergeTransactionRepository,
   private val staffUserAccountRepository: StaffUserAccountRepository,
   private val offenderFixedTermRecallRepository: OffenderFixedTermRecallRepository,
   private val sentencingAdjustmentService: SentencingAdjustmentService,
   private val linkCaseTxnRepository: LinkCaseTxnRepository,
   private val auditRepository: AuditRepository,
-  private val jdbcTemplate: JdbcTemplate,
+  private val imprisonmentStatusService: ImprisonmentStatusServiceProxy,
```
andymarke, line 113: By default this will continue to call Stored Procedure version

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderBookingRepository.kt
```diff
@@ -106,6 +96,11 @@ interface OffenderBookingRepository :
   @QueryHints(value = [QueryHint(name = "jakarta.persistence.lock.timeout", value = "2000")])
   @Query("from OffenderBooking ob where ob.bookingId = :bookingId")
   fun findByIdOrNullForUpdate(bookingId: Long): OffenderBooking?
+
+  @Lock(LockModeType.PESSIMISTIC_WRITE)
+  @QueryHints(value = [QueryHint(name = "jakarta.persistence.lock.timeout", value = "1000")])
+  @Query("from OffenderBooking ob where ob.bookingId = :bookingId")
+  fun findByIdWaitForLock(bookingId: Long): OffenderBooking
```
mikehalmamoj, line 103: Is this just the "cannot be null" version of the previous? Probably better if both had same naming convention, for update or wait for lock

mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderBookingRepository.kt
andymarke: I have deliberately chosen a "new" naming convention.

I found I was using it not to "get" the thing but to wait for a lock for the thing.
I did think about using a name like `acquireLockFor` but given it also returns the thing maybe something like `getByIdAndLock`. rather than `findByIdOrNullForUpdate` 

I vaguely remember back in the day  that `find` meant it might return something if it is found while  `get` will return something.

So I'm not too sure though. Let me have a think since this was deliberate .....

PR #1919, opened 2026-07-29
SDIT-4058: 🎨 Improve jpa mappings to reduce number of selects

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/coreperson/CorePersonService.kt
```diff
@@ -174,17 +174,13 @@ class CorePersonService(
...
     )
   }
 
-  fun getOffenderReligions(prisonNumber: String): List<OffenderBelief> {
-    val rootOffender =
-      offenderRepository.findRootByNomsId(prisonNumber) ?: throw NotFoundException("Offender not found $prisonNumber")
-    return offenderBeliefRepository.findByRootOffenderIdOrderByStartDateDescCreateDatetimeDesc(rootOffender.id)
-      .map { it.toBelief() }
-  }
+  fun getOffenderReligions(prisonNumber: String): List<OffenderBelief> = offenderBeliefRepository.findBeliefsByPrisonNumber(prisonNumber)
```
petergphillips, line 182: we were never doing anything with the not found since it was a can't happen in the first place

mikehalmamoj (dismissed):

mikehalmamoj (approved):

PR #1923, opened 2026-07-30
SDIT-4062 Sentence Term no longer calls imprisonment status

andymarke:
Term has no effect on status or main offence so no need to refresh at this point.

Alos renamed all repo methods that acquire lock to be explicit what it is doing

Also temp fix to deal with offender charge not found during imprisonment status calculation. This requires further work

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
@@ -708,7 +708,7 @@ class CourtSentencingService(
         "caseId" to caseId.toString(),
       )
 
-      courtEventRepository.findByIdOrNullForUpdate(eventId)?.also {
+      courtEventRepository.findByIdOrNullWaitForLock(eventId)?.also {
         val offenderCharges = it.courtEventCharges.map { cec -> cec.id.offenderCharge }
         case.courtEvents.remove(it)
         courtCaseRepository.saveAndFlush(case)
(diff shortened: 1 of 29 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
```diff
@@ -999,12 +999,7 @@ class CourtSentencingService(
       endDate = null,
       sentenceTermType = lookupSentenceTermType(termRequest.sentenceTermType),
     )
-    offenderSentenceTermRepository.saveAndFlush(term).also {
-      imprisonmentStatusService.recalculateImprisonmentStatusAndMainOffence(
```
andymarke, line 1003: main change - no need to call this just because a term has been added or update - it is not involved in the calculation

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/ImprisonmentStatusService.kt
```diff
@@ -109,10 +109,14 @@ class ImprisonmentStatusService(
   private fun updateMainOffence(booking: OffenderBooking, offenderChargeId: Long) {
     resetMainOffenceForOldCharge(booking, offenderChargeId)
 
-    val newMainOffenceCharge = booking.courtCases.flatMap { it.offenderCharges }.first { it.id == offenderChargeId }
+    val newMainOffenceCharge = booking.courtCases.flatMap { it.offenderCharges }.firstOrNull { it.id == offenderChargeId }
     // try to acquire lock
-    offenderChargeRepository.findByIdWaitForLock(newMainOffenceCharge.id)
-    newMainOffenceCharge.mostSeriousFlag = true
+    newMainOffenceCharge?.run {
```
andymarke, line 114: workaround for bug that needs investigating; this is find a charge that isn't in the hibernation entity yet - not sure why...

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderVisitBalanceRepository.kt
```diff
@@ -14,5 +14,5 @@ interface OffenderVisitBalanceRepository : CrudRepository<OffenderVisitBalance,
   @Lock(LockModeType.PESSIMISTIC_WRITE)
   @QueryHints(value = [QueryHint(name = "jakarta.persistence.lock.timeout", value = "20000")])
```
andymarke, line 15: @karenmillermoj any idea why this lock time out is 20 seconds?

mikehalmamoj (approved):

PR #1924, opened 2026-07-31
SDIT-4086 Provide separate endpoints for reconciliation calls

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/finance/PrisonerTransactionsResource.kt
```diff
@@ -147,6 +147,78 @@ class PrisonerTransactionsResource(
       transactionEntrySequence,
       pageSize,
     )
+
+  @GetMapping("/transactions/range/{entryDate}")
```
karenmillermoj, line 151: Reconciliation will do this call once rather than doing an extra call to the db each time.  There is currently no index on entryDate so can take a long time.  Will be speaking to classic team to see if it is possible to add an extra index.

petergphillips (approved):
LGTM

PR #1926, opened 2026-07-31
SDIT-4062 Move tests over to use new imprisonment status

andymarke:
Found issue with deleting breach hearings; the delete needed to happen before the imprisonment status is recalculated. I think this is the sympton of mixing updates via the hibernate entity and via HQL.

This is probably root cause of prod bug seen.


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/courtsentencing/CourtSentencingService.kt
```diff
@@ -1491,6 +1485,12 @@ class CourtSentencingService(
         }
       }
     }
```
andymarke, line 1487: Interesting bug here; calling the imprisonmentStatusService.recalculateImprisonmentStatusAndMainOffence before the courtEventRepository.delete(courtEvent) ended up with the delete not working. 

I suspect the issue is related to delete and offender booking implicit save conflicting - requires more investigation; but changing the order fixes but it indicates there is larger issue to be looked at next

steverendell (approved):
lgtm

PR #1930, opened 2026-08-04
SDIT-4088: Add JPA model for transfer schedules and wait lists

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderTransferScheduleWaitList.kt
```diff
@@ -0,0 +1,108 @@
...
+    value = [
+      JoinColumnOrFormula(
+        formula = JoinFormula(
+          value = "'${TransferScheduleStatus.TRN_SCH_STS}'",
+          referencedColumnName = "domain",
+        ),
+      ), JoinColumnOrFormula(column = JoinColumn(name = "WAIT_LIST_STATUS", referencedColumnName = "code")),
+    ],
+  )
+  var waitListStatus: TransferScheduleStatus,
```
andymarke, line 50: Checking that you check prod data has no rogue bad data that makes this fail?

andymarke: Same for the other ref data

mikehalmamoj: Yep I tried creating a minimal waitlist in dev and used that to inform the nullability of all fields.

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/transfers/schedule/TransferScheduleRepositoryIntTest.kt
```diff
@@ -0,0 +1,175 @@
...
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.Staff
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.repository.OffenderIndividualScheduleWaitListRepository
+import uk.gov.justice.digital.hmpps.nomisprisonerapi.jpa.repository.OffenderTransferScheduleOutRepository
+import java.time.LocalDate
+
+/**
+ * Temporary test to exercise the new JPA entities/repositories for transfer schedules and their wait lists,
+ * using the new test data builders. This will be superseded by resource-level tests in later PRs.
+ */
+class TransferScheduleRepositoryIntTest(
```
andymarke, line 25: I don't bother with this test; I rely on the resource test, so I would have a resource test ti get a schedule and proves the `save` and `get` works.

The updates will come out in your PUT resource tests.

But totally valid approach below

andymarke (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/transfers/schedule/TransferScheduleRepositoryIntTest.kt
mikehalmamoj: It's just an effort to keep PR size down - just test that the JPA is working, then resource tests coming soon.

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderTransferScheduleWaitList.kt
mikehalmamoj: Sorry I see what you mean now - and I do have 1 rogue transfer priority of PEN instead of 1, 2 or 3 🤔

PR #1931, opened 2026-08-05
SDIT-4088 Map transfer cancellation reason correctly

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/TransferCancellationReason.kt
```diff
@@ -0,0 +1,13 @@
...
+import jakarta.persistence.DiscriminatorValue
+import jakarta.persistence.Entity
+
+@Entity
+@DiscriminatorValue(TransferCancellationReason.TRANSFER_CANCELLATION_REASON)
+class TransferCancellationReason(code: String, description: String) : ReferenceCode(TRANSFER_CANCELLATION_REASON, code, description) {
+  companion object {
+    const val TRANSFER_CANCELLATION_REASON = "TRN_CNCL_RSN"
+    fun pk(code: String): Pk = Pk(TRANSFER_CANCELLATION_REASON, code)
+  }
```
mikehalmamoj, line 12: The main change is to model the transfer cancellation reason code which can be set on either a transfer schedule or waitlist

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderTransferScheduleWaitList.kt
```diff
@@ -63,7 +63,8 @@ class OffenderTransferScheduleWaitList(
       ), JoinColumnOrFormula(column = JoinColumn(name = "TRANSFER_PRIORITY", referencedColumnName = "code")),
     ],
   )
-  var transferPriority: TransferPriority,
+  @NotFound(action = NotFoundAction.IGNORE)
+  var transferPriority: TransferPriority? = null,
```
mikehalmamoj, line 67: Also added a change to handle a corrupt transfer priority as there is some live data that doesn't exist on the reference code table. As DPS require this data we'll just make the DTO value non-nullable and provide a default.

lbennett-moj (approved):

PR #1935, opened 2026-08-05
SDIT-4069: 🐛 GL transactions might be for different prisoner

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderSubAccountRepository.kt
```diff
@@ -47,7 +47,6 @@ interface OffenderSubAccountRepository : JpaRepository<OffenderSubAccount, Offen
           MAX(txn_entry_date) AS txn_entry_date,
           MAX(txn_entry_time) AS txn_entry_time
       FROM gl_transactions
-      WHERE offender_id = :offenderId
```
petergphillips, line 50: it is linked by `txn_id` anyway so doesn't need this clause

andymarke: So are you sure that doesn't have an performance implication - I can't tell since the SQL is complex

andymarke (approved):

PR #1939, opened 2026-08-06
SDIT-4132 Add force clone on court appearance

andymarke:
Where a future dated appearance is added onto a previous booking we ignore cloning since the parant appearance is typically added at the same time, and we don't want to clone twice.

This adds an override for repairs


petergphillips (approved):
LGTM

PR #1950, opened 2026-08-12
SDIT-4139 Add active request param when getting offenderIds by range

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/prisoners/PrisonersResource.kt
```diff
@@ -402,7 +402,10 @@ class PrisonersResource(private val prisonerService: PrisonerService) {
     @RequestParam(value = "toRootOffenderId", required = true)
     @Parameter(description = "Return prisoners with root offender id less than or equal to this value.")
     toRootOffenderId: Long,
-  ): List<PrisonNumberAndRootOffenderId> = prisonerService.findPrisonersInRange(fromRootOffenderId, toRootOffenderId)
+    @RequestParam(value = "active", required = true)
```
karenmillermoj: these changes will obv have a knock-on effect to from-nomis-migration as the active flag is mandatory (like the method it is based on)...so both services will need to be pushed out together

petergphillips: guess you could make it optional to start with and then change later to be mandatory? Not sure who else is calling this endpoint

petergphillips (dismissed):
LGTM

petergphillips (approved):
LGTM

PR #1956, opened 2026-08-14
SDIT-4142: ✨ Add in processing of merge event for core person

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderBelief.kt
```diff
@@ -27,11 +26,10 @@ data class OffenderBelief(
   var beliefId: Long = 0,
 
   @Column(name = "OFFENDER_BOOK_ID", nullable = false)
-  val booking: Long,
+  val bookingId: Long,
 
-  @ManyToOne(optional = false, fetch = FetchType.LAZY)
-  @JoinColumn(name = "ROOT_OFFENDER_ID", nullable = false)
-  var rootOffender: Offender,
+  @Column(name = "ROOT_OFFENDER_ID", nullable = false)
```
petergphillips, line 31: in the merge scenario the `rootOffender` will no longer exist in the database. We can't then have a JPA `Offender` link here so have to use the `id` instead

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/repository/OffenderBeliefRepository.kt
```diff
@@ -23,7 +23,8 @@ interface OffenderBeliefRepository : JpaRepository<OffenderBelief, Long> {
       select ob
       from OffenderBelief ob
       join fetch ob.beliefCode
-      where ob.rootOffender.nomsId = :prisonNumber
+      join fetch Offender o on ob.rootOffenderId = o.id
```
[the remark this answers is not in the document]
petergphillips: well spotted, will fix

src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/jpa/OffenderBelief.kt
richardpopple: Ah - so the OffenderBelief that we would be bringing back wouldn't have an Offender because they've been deleted.

richardpopple (approved):
LGTM

PR #1957, opened 2026-08-14
SDIT-4151 Return the deleted schedule event ID

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/movements/taps/application/TapApplicationService.kt
```diff
@@ -106,18 +106,22 @@ class TapApplicationService(
...
+            scheduleOutRepository.delete(it)
+            deletedEventId = it.eventId
+          }
         application.tapScheduleOuts.remove(schedule)
       }
     }
 
     return applicationRepository.save(application)
-      .let { UpsertTapApplicationResponse(offenderBooking.bookingId, it.tapApplicationId) }
+      .let { UpsertTapApplicationResponse(offenderBooking.bookingId, it.tapApplicationId, deletedEventId) }
```
mikehalmamoj, line 124: If we delete a schedule when updating a tap application we need to delete the schedule's mapping in the sync service

andymarke (approved):

